PR #3, opened 2022-03-01
Basic API server implementation

streamer45:
#### Summary

Slowly but steady, still not much to show, just setting up basic things.

Salient points:

- Main service skeleton
- Basic API server implementation (with TLS support). Implemented a sample `getVersion` handler
- Implemented basic TestHelper
- Added a document around project structure to be further augmented as the implementation takes place


cmd/rtcd/main.go
```diff
@@ -39,9 +41,25 @@ func main() {
 
 	logger.Info("rtcd: starting up")
 
+	service, err := service.New(cfg.Service, logger)
+	if err != nil {
+		logger.Error("rtcd: failed to create service", mlog.Err(err))
+		return
+	}
+
+	if err := service.Start(); err != nil {
```
cpoile, line 50: I know it's annoying, but are we ok with shadowing? (I am, personally, but depends on how we set up golangci-lint)

Makefile
```diff
@@ -1,3 +1,7 @@
+DEFAULT_GOARCH := $(shell go env GOARCH)
+BUILD_HASH = $(shell git rev-parse HEAD)
+LDFLAGS += -X "github.com/mattermost/rtcd/service.buildHash=$(BUILD_HASH)"
```
cpoile, line 3: I think we can simplify to:
```suggestion
LDFLAGS += -X "service.buildHash=$(BUILD_HASH)"
```

Makefile
```diff
@@ -1,3 +1,7 @@
+DEFAULT_GOARCH := $(shell go env GOARCH)
+BUILD_HASH = $(shell git rev-parse HEAD)
```
cpoile, line 2: It might make it easier if we use the short hash?
```suggestion
BUILD_HASH = $(shell git rev-parse --short HEAD)
```

Also, we could populate the version using the tags (which makes it easy to use the matterbuild plugin to set off a new version). Just a thought for the future, something like:
```
BUILD_TAG_LATEST = $(shell git describe --tags --match 'v*' --abbrev=0)
```

cpoile (approved):
Nice! Just little comments, nothing substantial (sorry). :)

cmd/rtcd/main.go
streamer45: I am generally okay with shadowing cause it keeps code simpler and I prefer to have a tight scope on the error variable, especially in cases where you don't necessarily return right away. By the way this reminds me I should have included `.golangci.yml`. Will make an extra push.

Makefile
streamer45: Interestingly I wasn't able to make it work that way and went with the full module version. I think the short form works if you are targeting the `main` package but not otherwise.

Makefile
streamer45: I was mainly being consistent with both MM and calls plugin as they both use the extended hash. Not a big deal though, we can revisit this if needed.

> Also, we could populate the version using the tags (which makes it easy to use the matterbuild plugin to set off a new version). 

Absolutely.

PR #4, opened 2022-03-04
WebSocket server implementation

streamer45:
#### Summary

PR implements a basic WebSocket server. It's not 100% complete but PR was getting larger so opted to send it now.

Some important things still missing:

- Documentation for exported methods and types.
- Connection timeout handler (ping-pong).
- More tests.

The API is rather simple, some highlights:

- Server implements the `http.Handler` to be easily hooked to the existing HTTP API server without need for coupling.
- Higher level won't be dealing with connections directly but through message passing (using connection ids).
- Two channels of communication, one to send and one to receive messages.

### DIFF
--- a/service/ws/server.go
+++ b/service/ws/server.go
@@ -1,0 +1,156 @@
+// Copyright (c) 2022-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+package ws
+
+import (
+	"fmt"
+	"net/http"
+	"sync"
+
+	"github.com/gorilla/websocket"
+	"github.com/mattermost/mattermost-server/v6/shared/mlog"
+)
+
+const (
+	sendChSize    = 256
+	receiveChSize = 256
+)
+
+type UpgradeCb func(connID string, w http.ResponseWriter, r *http.Request) error
+
+type Server struct {
+	cfg       Config
+	log       *mlog.Logger
+	conns     map[string]*conn
+	upgradeCb UpgradeCb
+	mut       sync.RWMutex
+	sendCh    chan Message
+	receiveCh chan Message
+}
+
+func NewServer(cfg Config, log *mlog.Logger, opts ...Option) (*Server, error) {
+	if err := cfg.IsValid(); err != nil {
+		return nil, fmt.Errorf("failed to validate config: %w", err)
+	}
+	s := &Server{
+		cfg:       cfg,
+		log:       log,
+		conns:     make(map[string]*conn),
+		sendCh:    make(chan Message, sendChSize),
+		receiveCh: make(chan Message, receiveChSize),
+	}
+
+	for _, opt := range opts {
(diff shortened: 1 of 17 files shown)


service/ws/server_test.go
```diff
@@ -0,0 +1,387 @@
...
+
+	s, addr, shutdown := setupServer(t, WithUpgradeCb(upgradeCb))
+	defer shutdown()
+	defer s.Close()
+
+	_, port, err := net.SplitHostPort(addr)
+	require.NoError(t, err)
+
+	u := url.URL{Scheme: "ws", Host: "localhost:" + port, Path: "/ws"}
+	c, _, err := websocket.DefaultDialer.Dial(u.String(), nil)
```
cpoile, line 101: cool!

service/ws/utils.go
```diff
@@ -0,0 +1,29 @@
...
+)
+
+const charset = "ybndrfg8ejkmcpqxot1uwisza345h769"
+
+var encoding = base32.NewEncoding(charset)
+
+// NewID is a globally unique identifier.  It is a [A-Z0-9] string 26
+// characters long.  It is a UUID version 4 Guid that is zbased32 encoded
+// with the padding stripped off.
+func newID() string {
```
cpoile, line 20: any reason we're not using `model.NewId()`?

cpoile (approved):
Awesome!

service/ws/utils.go
streamer45: The reason is I am a firm believer of [A little copying is better than a little dependency.](https://go-proverbs.github.io/)
We don't want to couple the service with server's `model` if we can avoid it cause they don't share anything.

service/ws/server.go
```diff
@@ -0,0 +1,156 @@
...
+			s.log.Error("ws read failed", mlog.Err(err))
+			break
+		}
+
+		var msgType MessageType
+		if mt == websocket.TextMessage {
+			msgType = TextMessage
+		} else if mt == websocket.BinaryMessage {
+			msgType = BinaryMessage
+		}
```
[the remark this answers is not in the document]
streamer45: We could, I just didn't know whether we had to move to a switch eventually so didn't really optimize :)

service/ws/server.go
```diff
@@ -0,0 +1,156 @@
...
+	for msg := range s.sendCh {
+		conn := s.getConn(msg.ConnID)
+		if conn == nil {
+			s.log.Error("failed to get conn for sending", mlog.String("connID", msg.ConnID))
+			continue
+		}
+
+		var msgType int
+		if msg.Type == TextMessage {
+			msgType = websocket.TextMessage
```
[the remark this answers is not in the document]
streamer45: This is still valuable feedback especially as I am big fan of code conciseness.

streamer45:
@cpoile I am going to merge this as I'll be addressing your feedback in the next PR.

PR #5, opened 2022-03-09
WebSocket client implementation

streamer45:
#### Summary

- Implemented WebSocket client.
- Added ping/pong support to server.
- Added some documentation.

This serves two main purposes:

- It makes testing the server easier and more complete.
- Client will be imported by the Calls plugin eventually when implementing the `plugin <-> rtcd` communication logic.

#### Next Steps

- Authentication service
- Connection management (at the service level)

### DIFF
--- a/service/ws/client.go
+++ b/service/ws/client.go
@@ -149,7 +149,6 @@ func (c *Client) ErrorCh() <-chan error {
 }
 
 // Close closes the underlying WebSocket connection.
-// Must be called once all sending operations are done and cannot be called more than once.
 func (c *Client) Close() error {
 	err := c.conn.close()
 	c.wg.Wait()
--- a/service/ws/client.go
+++ b/service/ws/client.go
@@ -1,0 +1,157 @@
+// Copyright (c) 2022-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+package ws
+
+import (
+	"fmt"
+	"net/http"
+	"sync"
+	"time"
+
+	"github.com/gorilla/websocket"
+)
+
+type Client struct {
+	cfg       ClientConfig
+	conn      *conn
+	sendCh    chan Message
+	receiveCh chan Message
+	errorCh   chan error
+	wg        sync.WaitGroup
+}
+
+// NewClient initializes and returns a new WebSocket client.
+func NewClient(cfg ClientConfig, opts ...ClientOption) (*Client, error) {
+	if err := cfg.IsValid(); err != nil {
+		return nil, fmt.Errorf("failed to validate config: %w", err)
+	}
+
+	header := http.Header{
+		"Authorization": []string{"Bearer " + cfg.AuthToken},
+	}
+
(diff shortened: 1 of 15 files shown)


service/ws/option.go
```diff
@@ -3,9 +3,12 @@
 
 package ws
 
-type Option func(s *Server) error
+type ServerOption func(s *Server) error
+type ClientOption func(c *Client) error
 
-func WithUpgradeCb(cb UpgradeCb) Option {
+// WithUpgradeCb lets the caller set an optional callback to be called prior to
+// performing the WebSocket upgrade.
```
cpoile, line 10: 😍 Thanks for the comment. 😄

service/ws/server.go
```diff
@@ -52,14 +56,32 @@ func NewServer(cfg Config, log *mlog.Logger, opts ...Option) (*Server, error) {
...
 	return s.sendCh
 }
 
+// ReceiveCh returns a channel that can be used to receive messages from ws connections.
 func (s *Server) ReceiveCh() <-chan Message {
 	return s.receiveCh
 }
 
+// Close stops the websocket server and closes all the ws connections.
+// Must be called once all senders are done and cannot be called more than once.
```
cpoile, line 70: also sync.Once -able?

service/ws/client.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+// ErrorCh returns a channel that is used to receive client errors
+// asynchronously.
+func (c *Client) ErrorCh() <-chan error {
+	return c.errorCh
+}
+
+// Close closes the underlying WebSocket connection.
+// Must be called once all sending operations are done and cannot be called more than once.
```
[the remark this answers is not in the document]
streamer45: Actually looking at this I think it's totally safe to be called more than once as we are not closing any channel.

service/ws/config.go
```diff
@@ -5,19 +5,61 @@ package ws
...
+	if c.URL == "" {
+		return fmt.Errorf("invalid URL value: should not be empty")
+	}
+
+	if !strings.HasPrefix(c.URL, "ws://") && !strings.HasPrefix(c.URL, "wss://") {
+		return fmt.Errorf(`invalid URL value: should start with "ws://" or "wss://"`)
+	}
+
+	if c.ConnID != "" && len(c.ConnID) != 26 {
+		return fmt.Errorf("invalid ConnID value: too short")
```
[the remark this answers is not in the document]
streamer45: I am clearly not native speaker, allow me to rephrase :stuck_out_tongue:

service/ws/server.go
streamer45: It would but the assumption here is that the caller should not close it twice and if it does there's a problem somewhere. If we make it silently go through we'd never know. At this stage I'd prefer it to panic/fail test if we screw up something higher up. Of course if at some point we come up with a use-case for potentially closing it twice then by all means we should make use of something like `sync.Once`.

PR #6, opened 2022-03-15
Auth implementation

streamer45:
#### Summary

PR implements the main authentication logic for the service. Key points:

- We use an embedded persistent K/V store to save credentials. This may be temporary and I can see us moving to something centralized as the service evolves (e.g. multiple instances). We can support various store implementations behind a common interface based on the need.
- Implemented simple auth service to register and authenticate clients.
- Implemented WebSocket authentication handler using basic auth.

The expected authentication flow can be found in [docs/security.md](https://github.com/mattermost/rtcd/compare/auth?expand=1#diff-2c506cc756a5333f8fb5acb850d443980706602048f9ddb3dc515abacbfbce6a)

### DIFF
--- a/service/auth.go
+++ b/service/auth.go
@@ -4,9 +4,56 @@
 package service
 
 import (
+	"encoding/json"
+	"fmt"
 	"net/http"
+
+	"github.com/mattermost/mattermost-server/v6/shared/mlog"
 )
 
 func (s *Service) wsAuthHandler(connID string, w http.ResponseWriter, r *http.Request) error {
+	clientID, authKey, ok := r.BasicAuth()
+	if !ok {
+		w.WriteHeader(http.StatusUnauthorized)
+		return fmt.Errorf("authentication failed: invalid auth header")
+	}
+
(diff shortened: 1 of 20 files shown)


service/auth.go
```diff
@@ -4,9 +4,56 @@
...
+
+	if err := s.auth.Authenticate(clientID, authKey); err != nil {
+		w.WriteHeader(http.StatusUnauthorized)
+		return fmt.Errorf("authentication failed: %w", err)
+	}
+
 	return nil
 }
+
+func (s *Service) registerClient(w http.ResponseWriter, req *http.Request) {
```
streamer45, line 29: The `/register` API endpoint doesn't currently require any authentication, which means clients can freely register themselves. This could potentially change or be configured if needed.

cpoile (approved):
I can't speak to the security aspects (not qualified), but otherwise looks great! Love the tests. That's a lot of work, and it's a good roll-model for PRs.

oh6hay:
LGTM for now for now and for testing :+1:

PR #7, opened 2022-04-06
RTC implementation

streamer45:
#### Summary

PR implements the RTC server. Most of the WebRTC logic (SFU implementation) has been ported from Calls plugin. This was done in a such a way that the plugin can now directly import the `rtc` package and work just like before (see https://github.com/mattermost/mattermost-plugin-calls/pull/40).

Implementation wise I kept the pattern very similar to previous services, relying on message passing when feasible to easily allow for usage over network as well. The next step in fact will be to make this work through the `service` package and enable the `rtcd` service from the plugin.

I added some tests and documentation to hopefully make the overall functioning clearer to understand. These are not comprehensive yet, something to improve in a later refinement.


docs/implementation.md
```diff
@@ -0,0 +1,89 @@
...
+
+```mermaid
+flowchart TD
+service-->api[api\nHTTP API Server]
+service-->ws[ws\nWebSocket Server]
+service-->rtc[rtc\nWebRTC SFU]
+service-->store[store\nK/V Store]
+service-->auth[auth\nAuthentication Service]
+auth-->store
+```
```
cpoile, line 18: This is really cool, nicely done.

docs/implementation.md
```diff
@@ -0,0 +1,89 @@
...
+The `auth` packages implements a simple authentication service to register, unregister and authenticate clients.
+
+### `store`
+
+Store provides a basic interface to key-value store. Its main implementation is currently based on [bitcask](https://git.mills.io/prologic/bitcask), a persistent embedded key-value store.
+
+## Sample deployment
+
+```mermaid
+flowchart TB
```
cpoile, line 48: now you're just showing off 😝

service/rtc/msg.go
```diff
@@ -0,0 +1,58 @@
...
+	"encoding/json"
+	"fmt"
+
+	"github.com/pion/webrtc/v3"
+)
+
+type MessageType int
+
+const (
+	ICEMessage MessageType = iota + 1
```
cpoile, line 16: just curious, why +1? Is 0 reserved (or implicitly means nil)?

service/service.go
```diff
@@ -72,17 +79,47 @@ func New(cfg Config, log *mlog.Logger) (*Service, error) {
...
+			switch msg.Type {
+			case ws.OpenMessage:
+				s.log.Debug("connect", mlog.String("connID", msg.ConnID))
+			case ws.CloseMessage:
+				s.log.Debug("disconnect", mlog.String("connID", msg.ConnID))
+			case ws.TextMessage:
+				s.log.Warn("unexpected text message", mlog.String("connID", msg.ConnID))
+			case ws.BinaryMessage:
+				var clientMsg ClientMessage
+				if err := clientMsg.Unpack(msg.Data); err != nil {
```
cpoile, line 100: Is this going to be a hot path? Should we unpack in a go routine? Or is it negligible?

cpoile (approved):
Really nice, thanks Claudio!

service/service.go
streamer45: Only time (and data) will tell us. At the moment we have a very low rate of events per second, even when lots of users are joining calls. Keeping it simple for now until further optimization is required.

service/rtc/msg.go
streamer45: 0 being the default value for a integer variable I tend to avoid using it so that the event has to be explicitly set.

PR #9, opened 2022-04-12
Service implementation

streamer45:
#### Summary

This is the last piece of implementation that lets Calls plugin instances successfully interface with the service over network.
PR implements a `service.Client` to use the provided HTTP/WebSocket API.

Bear in mind that the sole purpose of this PR is establishing main functionality (getting calls to work). There isn't any real improvement (e.g. reconnection support) just yet. Also, APIs are very likely to break for some time still as we are trying to come up with a proper solution on the security front.

Companion PR is https://github.com/mattermost/mattermost-plugin-calls/pull/41

### DIFF
--- a/service/rtc/server.go
+++ b/service/rtc/server.go
@@ -28,7 +28,8 @@ type Server struct {
 	log     mlog.LoggerIFace
 	metrics Metrics
 
-	groups map[string]*group
+	groups   map[string]*group
+	sessions map[string]SessionConfig
 
 	udpConn *net.UDPConn
 	udpMux  ice.UDPMux
@@ -55,6 +56,7 @@ func NewServer(cfg ServerConfig, log mlog.LoggerIFace, metrics Metrics) (*Server
 		log:       log,
 		metrics:   metrics,
 		groups:    map[string]*group{},
+		sessions:  map[string]SessionConfig{},
 		sendCh:    make(chan Message, msgChSize),
 		receiveCh: make(chan Message, msgChSize),
 	}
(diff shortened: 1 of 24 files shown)


service/rtc/server.go
```diff
@@ -147,25 +154,37 @@ func (s *Server) Stop() error {
 func (s *Server) msgReader() {
 	for msg := range s.sendCh {
 		if err := msg.IsValid(); err != nil {
-			s.log.Error("invalid message", mlog.Err(err))
+			s.log.Error("invalid message", mlog.Err(err), mlog.Int("msgType", int(msg.Type)))
+			continue
+		}
+
+		s.mut.RLock()
```
cpoile, line 161: I wonder if it would be better to lock the maps individually, instead of the entire struct? That might cause unnecessary locking for something that's going to happen quite often, right?

service/ws/conn.go
```diff
@@ -72,3 +85,15 @@ func (s *Server) getConns() []*conn {
 	}
 	return conns
 }
+
+func (s *Server) getClientConns(clientID string) []*conn {
+	s.mut.RLock()
+	defer s.mut.RUnlock()
```
cpoile, line 91: Maybe a similar issue here (and above)? But I'm not sure how hot this path is, so maybe it's not as important?

cpoile (approved):
Awesome! I can't think of much to add, amazing work!

service/rtc/server.go
streamer45: It helps that this is read-locking which means it will only be waiting if at the same time a new session is being added which is a very fast operation to start with. The mutex per map optimization is totally valid but I am very cautious to add more complexity unless it's really needed. We will monitor everything of course, taking CPU profiles and such and if lock contention becomes an issue I am happy to improve further.

PR #10, opened 2022-04-14
Feature: Introduction of Dockerfile 

phoinixgrr:
#### Summary

This introduces a multi-stage Dockerfile that builds the rtcd artifact.

#### Ticket Link

https://mattermost.atlassian.net/browse/DOPS-917

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,12 +3,11 @@ BUILD_HASH = $(shell git rev-parse HEAD)
 LDFLAGS += -X "github.com/mattermost/rtcd/service.buildHash=$(BUILD_HASH)"
 
 ## Check go mod files consistency
-.PHONY: gomod-check
+.PHONY: gomod-check check-style golangci-lint test build clean
 gomod-check:
 	@echo Checking go mod files consistency
 	go mod tidy -v && git --no-pager diff --exit-code go.mod go.sum || (echo "Please run \"go mod tidy\" and commit the changes in go.mod and go.sum." && exit 1)
 
-.PHONY: check-style
 check-style: golangci-lint gomod-check
 	@echo Checking for style guide compliance
 
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -1,0 +1,24 @@
+# This dockerfile is used to build Mattermost rtcd
+# A multi stage build, with golang:1.17.9-alpine used as a builder
+# and gcr.io/distroless/static as runner 
+FROM golang:1.17.9-alpine@sha256:5c2fcfeb58ad9d4948d94e7b2d0432a9bc38bee0f8dfb41d383f38a18b75c38d as builder
+RUN apk update && apk add make
+
+# Setup Golang directories structure and compile
+RUN mkdir -p /rtcd
+ADD . /rtcd
+WORKDIR /rtcd
+RUN make build
+
+# Shrink final image since we only need the rtcd binary
+# and use distroless container image as runner for security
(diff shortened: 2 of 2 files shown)


pfltdv (approved):
🏆 Nice work.

streamer45 (commented):
Thanks @phoinixgrr , looks great :muscle: , I have only one question:

I see that running this without any override returns some permission errors related to the log file:
```
can't open new logfile: open rtcd.log: permission denied
```

Is there anything we can do to avoid this? I assume the running user should have permissions to write logs although I understand that in reality we would probably save them to an external path (e.g. volume).

phoinixgrr:
> Thanks @phoinixgrr , looks great 💪 , I have only one question:
> 
> I see that running this without any override returns some permission errors related to the log file:
> 
> ```
> can't open new logfile: open rtcd.log: permission denied
> ```
> 
> Is there anything we can do to avoid this? I assume the running user should have permissions to write logs although I understand that in reality we would probably save them to an external path (e.g. volume).

Hey @streamer45,

As best practice I would recommend refraining from saving logs into the container since we need the container to be stateless. What is recommended as best practise, is to push the logs in standard output. Thus they can be easily parsed by log aggregation system such as Grafana Loki. 

Does this makes sense?

streamer45:
> As best practice I would recommend refraining from saving logs into the container since we need the container to be stateless. What is recommended as best practise, is to push the logs in standard output. Thus they can be easily parsed by log aggregation system such as Grafana Loki.
> 
> Does this makes sense?

Alright, then the suggestion is to disable file logging when running in a container. Sounds good :+1:

streamer45 (approved):

spirosoik (commented):
minor small change @phoinixgrr . By my experience we are trying to follow the `Standard Go Project Layout` where the `Dockerfile` lives under `/build`.

A bit of context here https://github.com/golang-standards/project-layout .

An example by MM server https://github.com/mattermost/mattermost-server/tree/master/build

phoinixgrr:
@spirosoik  I have pushed the required changes. Let me know if I got this right 

Also pushed some fixes to the `Makefile` with https://github.com/mattermost/rtcd/pull/10/commits/f6a7abe2bac7d98022246cdbbb03a9bfefcb1d83 
Thanks for Re-reviewing @streamer45  @pfltdv

Makefile
```diff
@@ -3,7 +3,7 @@ BUILD_HASH = $(shell git rev-parse HEAD)
 LDFLAGS += -X "github.com/mattermost/rtcd/service.buildHash=$(BUILD_HASH)"
 
 ## Check go mod files consistency
-.PHONY: gomod-check
+.PHONY: gomod-check check-style golangci-lint test build clean
```
pfltdv, line 6: `.PHONY: check-style` is defined at `check-style` task at line 11. Is it possible to remove that or define `.PHONY` per task?

phoinixgrr: Missed that. Rebased and fixed!

streamer45 (approved):
Awesome :)

agnivade:
@spirosoik - 

> minor small change @phoinixgrr . By my experience we are trying to follow the Standard Go Project Layout where the Dockerfile lives under /build.

> A bit of context here https://github.com/golang-standards/project-layout 

We are certainly not using this. This is not recommended by the Go team and some recommendations go against Go best practices.

It's obviously fine to to place Dockerfiles under `/build`. But as general guidelines, we don't follow https://github.com/golang-standards/project-layout. Our recommended project layout is here: https://developers.mattermost.com/contribute/server/style-guide/#project-layout.

spirosoik:
@agnivade thanks for the information, I was actually mentioning only the `/build` part but thanks for sharing

PR #11, opened 2022-04-14
[MM-43391] Use defaults if config file is missing

streamer45:
#### Summary

Letting the service use defaults if the config file is missing. This makes the service run even if not provided with a config file and removes the necessity of including such file in the docker image.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-43391

### DIFF
--- a/cmd/rtcd/config.go
+++ b/cmd/rtcd/config.go
@@ -4,7 +4,10 @@
 package main
 
 import (
+	"errors"
 	"fmt"
+	"log"
+	"os"
 
 	"github.com/mattermost/rtcd/logger"
 	"github.com/mattermost/rtcd/service"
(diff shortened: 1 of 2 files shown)


phoinixgrr (approved):
LGTM 🎖️ 

You think it would be a good idea to also document the related configuration environment variables?

spirosoik (approved):
great, thanks @streamer45

streamer45:
> LGTM medal_military
> 
> You think it would be a good idea to also document the related configuration environment variables?

I think that's a great idea but I'd like to find a way of achieving that without requiring us to update it every time we make a change. I created [MM-43394](https://mattermost.atlassian.net/browse/MM-43394) as a follow-up ticket to address this.

PR #12, opened 2022-04-14
[MM-43394] Document config environment overrides

streamer45:
#### Summary

PR simplifies the configuration by removing an unnecessary level of nesting. This means the logger is now created by the service itself which I think it's acceptable.

As a bonus I've added a new `make doc` command to generate a [list of environment config overrides](https://github.com/mattermost/rtcd/blob/MM-43394/docs/env_config.md). Ideally I'd like to improve it and also document each config setting using tags but that requires far more effort so I am happy to have something rather than nothing for the time being.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-43394

/cc @phoinixgrr


cpoile (approved):
Cool! Nice user-friendly use of tabwriter. :)

phoinixgrr:
🎖️  🏅  🏆

PR #16, opened 2022-04-19
High concurrency fixes

streamer45:
#### Summary

Addressing two issues that were found during initial performance testing.

- Making `rtc.Server.InitSession()` partially synchronous to avoid a race condition in which events could be processed before the connection they targeted was fully initiated in the state.
- Adding a map to the service to keep track of which MM instance the end users connections (session ids) belong to. This is to avoid WebRTC negotiation errors caused by out of order delivery of websocket messages.

### DIFF
--- a/service/rtc/sfu.go
+++ b/service/rtc/sfu.go
@@ -316,31 +316,36 @@ func (s *Server) InitSession(cfg SessionConfig) error {
 		}
 	})
 
-	select {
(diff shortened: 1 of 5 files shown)


streamer45:
@cpoile To your excellent point earlier on, I verified that in case of a reconnection to a different HA node the connection is created from scratch as we are not supporting this case yet.

cpoile (approved):
Awesome!

PR #18, opened 2022-04-27
Performance improvements

streamer45:
#### Summary

PR addresses two issues discovered during initial performance testing.

1. We add a callback to be run on connection close and also attempt to close the connection on the rtc side as well when failure occurs. This prevents possible resource leaks due to missing events on the websocket channel with the plugin (e.g. mattermost restarts as a user disconnects).
2. We listen to the same addr/port multiple times (once per CPU available) to allow for better receive throughput. This optimization has showed the best improvements during testing. 

Implementation wise I originally hacked the [`pion.ice`](https://pkg.go.dev/github.com/pion/ice/v2) code as it made it for a much  simpler solution which also provided higher read/write concurrency. Unfortunately that required breaking changes plus a need to port everything upstream so went with a less optimal approach by implementing a `multiConn` object to do everything on our side. So we are still passing a single `net.PacketConn` but in reality this is wrapping multiple connections. Limitations with the current approach:

- Writes are still happening from a single goroutine.
- While reads are concurrent there's still a bottleneck of a single caller as all results need to go through the same channel.
- We have to copy data buffers.

That said, I load-tested both approaches and overall there wasn't any noticeable difference but I'd fully expect the latter to struggle sooner when reaching the instance limits.

Fortunately [`ice.UDPMux`](https://pkg.go.dev/github.com/pion/ice/v2#UDPMux) is an interface so we still have the option to provide our own implementation which may be the best solution from a performance perspective although it will require more implementation effort.

### DIFF
--- a/service/rtc/multi_conn.go
+++ b/service/rtc/multi_conn.go
@@ -1,0 +1,131 @@
+// Copyright (c) 2022-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+package rtc
+
+import (
+	"errors"
+	"net"
+	"os"
+	"sync"
+	"sync/atomic"
+	"time"
+)
+
+const (
+	receiveMTU = 8192
+)
+
+type multiConn struct {
+	conns        []net.PacketConn
+	addr         net.Addr
+	readResultCh chan readResult
+	closeCh      chan struct{}
+	bufPool      *sync.Pool
+	counter      uint64
+	wg           sync.WaitGroup
+}
+
+type readResult struct {
+	n    int
+	addr net.Addr
+	err  error
+	buf  []byte
+}
+
+func newMultiConn(conns []net.PacketConn) (*multiConn, error) {
+	if len(conns) == 0 {
+		return nil, errors.New("conns should not be empty")
+	}
+	for _, conn := range conns {
+		if conn == nil {
+			return nil, errors.New("invalid nil conn")
+		}
+	}
+	var mc multiConn
+	mc.conns = conns
+	mc.addr = conns[0].LocalAddr()
+	mc.readResultCh = make(chan readResult)
+	mc.closeCh = make(chan struct{})
+	mc.bufPool = &sync.Pool{
+		New: func() interface{} {
+			return make([]byte, receiveMTU)
+		},
+	}
+	mc.wg.Add(len(conns))
+	for _, conn := range conns {
+		go mc.reader(conn)
+	}
+	return &mc, nil
+}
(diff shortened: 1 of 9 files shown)


service/rtc/multi_conn.go
```diff
@@ -0,0 +1,131 @@
...
+	res := <-mc.readResultCh
+	copy(p, res.buf[:res.n])
+	mc.bufPool.Put(res.buf)
+	return res.n, res.addr, res.err
+}
+
+func (mc *multiConn) WriteTo(p []byte, addr net.Addr) (n int, err error) {
+	// Simple round-robin to equally distribute the writes among the connections.
+	idx := (atomic.AddUint64(&mc.counter, 1) - 1) % uint64(len(mc.conns))
+	return mc.conns[idx].WriteTo(p, addr)
```
streamer45, line 91: I tested random as well but clearly it was much worse than atomic + modulo. I think round-robin works particularly well in this case.

cpoile (approved):
Looks good to me, but tbh I'm a little beyond my depth on the multi_conn, sorry... 
I keep wondering (for us to talk about in the future, not for this PR)--since go's channels are known to be much slower than the lower level concurrency primitives, if we might think about manually managing udp traffic. A lot more work, more error prone, not as convenient for sure, but if performance becomes a bottleneck it might be a possible solution.

streamer45:
> Looks good to me, but tbh I'm a little beyond my depth on the multi_conn, sorry... I keep wondering (for us to talk about in the future, not for this PR)--since go's channels are known to be much slower than the lower level concurrency primitives, if we might think about manually managing udp traffic. A lot more work, more error prone, not as convenient for sure, but if performance becomes a bottleneck it might be a possible solution.

On a theoretical level you are spot on. My more practical take on this is that if we are willing to leave go channels behind due to perf reasons we might as well leave Golang. More generally I think the real solution is to make this scale horizontally as our goal is not to have the most efficient SFU implementation ever. In other words, at our current scale I think it's not the best use of our time to over optimize.

agnivade:
> since go's channels are known to be much slower than the lower level concurrency primitives, 

This is similar to saying C is faster than Go. There is obviously a non-zero cost to using channels but I am fairly confident there will always be some other bottleneck in the code. If you ever feel that the code is optimized to a level that channels are turning out to be a bottleneck, then I'd be curious to look into your profiles to see the proof.

In short, I am yet to see real-life code (not artificially written microbenchmarks) so well-optimized, that channels become a bottleneck. So I won't worry about using channels just yet. :)

PR #19, opened 2022-04-27
Fix env vars document generation

streamer45:
#### Summary

We empty the file prior to writing or it gets overwritten potentially causing some issues.


cpoile (approved):
👍

PR #24, opened 2022-05-04
[MM-43508] Implement basic reconnection support for rtcd client

streamer45:
#### Summary

PR adds basic reconnection support to the `rtcd` client. 

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-43508

### DIFF
--- a/service/client.go
+++ b/service/client.go
@@ -18,25 +18,35 @@ import (
 )
 
 const (
-	msgChSize = 64
+	msgChSize                = 64
+	maxReconnectInterval     = 30 * time.Second
+	defaultReconnectInterval = 2 * time.Second
 )
 
 type Client struct {
-	cfg *ClientConfig
+	cfg    *ClientConfig
+	connID string
 
-	httpClient *http.Client
-	wsClient   *ws.Client
-	receiveCh  chan ClientMessage
-	errorCh    chan error
+	httpClient  *http.Client
+	wsClient    *ws.Client
+	receiveCh   chan ClientMessage
+	errorCh     chan error
+	reconnectCb ClientReconnectCb
+	closed      bool
+
+	mut sync.RWMutex
+	wg  sync.WaitGroup
 }
 
-func NewClient(cfg ClientConfig) (*Client, error) {
+func NewClient(cfg ClientConfig, opts ...ClientOption) (*Client, error) {
 	var c Client
 
 	if err := cfg.Parse(); err != nil {
 		return nil, fmt.Errorf("failed to parse config: %w", err)
 	}
 	c.cfg = &cfg
+	c.receiveCh = make(chan ClientMessage, msgChSize)
+	c.errorCh = make(chan error)
 
 	transport := &http.Transport{
 		Proxy: http.ProxyFromEnvironment,
@@ -55,6 +65,12 @@ func NewClient(cfg ClientConfig) (*Client, error) {
 	}
 
 	c.httpClient = &http.Client{Transport: transport}
+
+	for _, opt := range opts {
+		if err := opt(&c); err != nil {
+			return nil, fmt.Errorf("failed to apply option: %w", err)
+		}
+	}
 
 	return &c, nil
 }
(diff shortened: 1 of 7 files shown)


service/option.go
```diff
@@ -0,0 +1,16 @@
...
+type ClientReconnectCb func(c *Client, attempt int) error
+
+// WithClientReconnectCb lets the caller set an optional callback to be called prior to
+// performing a WebSocket reconnection.
+func WithClientReconnectCb(cb ClientReconnectCb) ClientOption {
+	return func(c *Client) error {
+		c.reconnectCb = cb
+		return nil
+	}
+}
```
cpoile, line 16: tbh, this is bending my mind a little -- it's a modifier that takes a cb, and returns a ClientOption, which attaches the cb to the client.  I feel like I just want a method on the client `RegisterReconnectCb`. But maybe you need this way instead?

service/client.go
```diff
@@ -160,28 +183,40 @@ func (c *Client) Connect() error {
...
-		wg.Wait()
-		close(c.errorCh)
 	}()
 
 	return nil
 }
 
 func (c *Client) Send(msg ClientMessage) error {
+	c.mut.RLock()
+	defer c.mut.RUnlock()
```
cpoile, line 214: Is there a reason these are added now?
Or, I think I'm not understanding what the `c.mut` is specifically locking... Is it the `c.closed` bool?

service/client.go
```diff
@@ -234,10 +285,62 @@ func (c *Client) msgReader() {
...
+
+func (c *Client) reconnectHandler() {
+	var attempt int
+	var waitTime time.Duration
+	for {
+		attempt++
+		if waitTime < maxReconnectInterval {
+			waitTime += c.cfg.ReconnectInterval
+		}
+		time.Sleep(waitTime)
```
cpoile, line 313: Does this mean we'll wait a minimum 2 seconds after being disconnected?  I wonder if that's excessive, at least for the first attempt? Should we try to reconnect almost immediately?

service/client_test.go
```diff
@@ -300,21 +317,226 @@ func TestClientReceive(t *testing.T) {
...
-		err = th.srvc.wsServer.Send(ws.Message{Type: ws.BinaryMessage, Data: data})
+		err = th.srvc.wsServer.Send(ws.Message{Type: ws.BinaryMessage, Data: data, ConnID: msgData["connID"], ClientID: clientID})
 		require.NoError(t, err)
 	}
 
-	err = c.Close()
+	wg.Wait()
+}
+
+func TestClientReconnect(t *testing.T) {
```
cpoile, line 341: nice tests!

cpoile (commented):
This is awesome.  A little complicated in parts, which are what my comments are about. Nothing really blocking, but would love to know more about the questions. Thanks Claudio!

service/option.go
streamer45: This approach is rather common throughout the codebase (e.g. [here](https://github.com/mattermost/mattermost-server/blob/master/app/options.go)) which is probably the main reason I went for it. It's a clever way to avoid breaking the API and still allow for future extensibility.

We don't strictly need to do this, obviously this could be implemented in a number of other ways but I think it's simple enough once understood and it comes with some bonuses as mentioned before plus doing it this way we ensure the cb is set at initialization time so we don't have to worry about races.

service/client.go
streamer45: Well, the reason is that now we are touching state from multiple goroutines and need to be extra careful. Of course the logic inevitably becomes more complex.

In fact, the reconnect handler runs in the reader goroutine and could trigger concurrently with any other operation (e.g. Close).

Once you start allowing for concurrency then there's no turning back really so I had to also make sure every method was safe to be called concurrently with eachother

service/client.go
streamer45: Attempting a reconnect in the same millisecond is unlikely to work given either the network is down or the service restarted/crashed so I think some wait time makes sense. I made the interval configurable so we can definitely lower it on the other side if we want to.

service/client.go
cpoile: Ok, gotcha -- I missed that we were touching from multiple goroutines. Now it makes sense (here and all the other places).

cpoile (approved):
Nice! 👍

PR #27, opened 2022-05-09
[MM-44062] Use STUN to find public IP

streamer45:
#### Summary

We are moving to this side the STUN logic to figure out an instance's own public IP address which will trigger only in case the `ICEHostOverride` is empty and ICE servers are provided.

This is especially helpful for Cloud operations now that we have simplified the networking design, with pods having public IP addresses.

Big thanks to @angeloskyratzakos for quickly validating that this approach can work.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-44062


cpoile (approved):
👍

PR #28, opened 2022-05-10
Change registration flow to require client provided auth key

streamer45:
#### Summary

PR changes the registration flow from server-generated to client-provided key. This lets us simplify quite a lot on the plugin (client) side especially in preparation to supporting multiple rtcd servers at once.

The client-provided auth key will use the same generation function so that we guarantee the same length/entropy.

### DIFF
--- a/docs/security.md
+++ b/docs/security.md
@@ -8,10 +8,9 @@ The authentication flow is as follows:
 
 #### Registration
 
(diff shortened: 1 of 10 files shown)


cpoile (approved):
Looks good from my end (can't speak to the security aspect of it). Thanks @streamer45!

oh6hay (approved):
LGTM!

PR #33, opened 2022-05-12
[fix] Makefile no need to run go-build before docker-build

phoinixgrr:
#### Summary

No need to run go-build before docker-build. Build happens within the Dockerfile
https://github.com/mattermost/rtcd/blob/master/build/Dockerfile#L13

#### Ticket Link


pfltdv (approved):
Nice catch!

PR #35, opened 2022-05-13
[MM-44126] Implement `/stats` API endpoint

streamer45:
#### Summary

PR adds a new `/stats` API endpoint to let clients fetch some statistics such as number of active sessions and calls. This will be particularly useful on the plugin side to determine which rtcd server to pick when load balancing between multiple hosts.

Returned stats are implicitly scoped by clientID as we leverage the existing prometheus metrics but I could see us adding global counters as well, just needs some more work.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-44126

### DIFF
--- a/service/auth.go
+++ b/service/auth.go
@@ -13,11 +13,7 @@ import (
 
 func (s *Service) authHandler(w http.ResponseWriter, r *http.Request) (clientID string, code int, err error) {
 	defer func() {
-		data := &httpData{
-			reqData: map[string]string{},
-			resData: map[string]string{},
-		}
-
+		data := newHTTPData()
 		data.code = code
 		if err != nil {
 			data.err = err.Error()
@@ -54,10 +50,7 @@ func (s *Service) registerClient(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	data := &httpData{
-		reqData: map[string]string{},
-		resData: map[string]string{},
-	}
+	data := newHTTPData()
 	defer s.httpAudit("registerClient", data, w, r)
 
 	if !s.cfg.API.Security.AllowSelfRegistration {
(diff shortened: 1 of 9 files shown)


service/perf/metrics.go
```diff
@@ -71,10 +72,21 @@ func NewMetrics(namespace string, registry *prometheus.Registry) *Metrics {
 			Name:      "sessions_total",
 			Help:      "Total number of active RTC sessions",
 		},
-		[]string{"groupID", "callID"},
+		[]string{"groupID"},
```
streamer45, line 75: Not sure why I added `callID` in the first place as it's not a very useful label. Now only considering groups (aka clients).

service/auth.go
```diff
@@ -13,11 +13,7 @@ import (
 
 func (s *Service) authHandler(w http.ResponseWriter, r *http.Request) (clientID string, code int, err error) {
 	defer func() {
-		data := &httpData{
-			reqData: map[string]string{},
-			resData: map[string]string{},
-		}
-
+		data := newHTTPData()
```
cpoile, line 16: Cleaner, nice.

cpoile (approved):
Do we need programmatic access to global counters? We could aggregate on the collector side until then, right?

streamer45:
> Do we need programmatic access to global counters? We could aggregate on the collector side until then, right?

Global counters (global to a given rtcd instance that is) would make the most sense as the metric to use when load balancing given global utilization is what ultimately counts towards performance. Separate clients/installations could have drastically different resource usages so with just these metrics exposed a client would be guessing purely based on its own usage. To improve this I can see two options:

- Leverage the existing counters and return the sum of all the metrics for each registered client.
- Implement global counters.

PR #36, opened 2022-05-16
Makefile Essential improvements 

phoinixgrr:
#### Summary

Makefile Essential improvements 

- [feat] Makefile tag latest when building from master (protected branch) 
- [feat] Makefile extract GO_VERSION from go.mod 
- [fix] Makefile disable volume mounts for cosign commands 
- [fix] No need to perform go-mod-check on lint
- [chore] makefile expose GOLANG as LDFLAG

Review by commit is recommended

#### Ticket Link

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,12 +51,17 @@ GO_TEST_OPTS                 += -mod=readonly -failfast -race
 GO_OUT_BIN_DIR               := ./dist
 
 ## General Variables
+# Branch Variables
+PROTECTED_BRANCH := master
+CURRENT_BRANCH   := $(shell git rev-parse --abbrev-ref HEAD)
 # Use repository name as application name
 APP_NAME    := $(shell basename -s .git `git config --get remote.origin.url`)
 # Get current commit
 APP_COMMIT  := $(shell git log --pretty=format:'%h' -n 1)
-# Check if we are in a release tag, if yes use it as app version.
+# Check if we are in a release tag, if yes use it as app version, else use dev-sha as app version
 APP_VERSION := $(shell git describe --abbrev=0 --exact-match --tags > /dev/null 2>&1 || echo dev-$(APP_COMMIT))
+# Check if we are in protected branch, if yes use latest as app version.
+APP_VERSION := $(shell if [ $(PROTECTED_BRANCH) = $(CURRENT_BRANCH) ]; then echo latest; else echo $(APP_VERSION) ; fi)
 # Get current date and format like: 2022-04-27 11:32
 BUILD_DATE  := $(shell date +%Y-%m-%d\ %H:%M)
 
@@ -115,7 +120,7 @@ test: go-test ## to test all
 
 .PHONY: docker-build
 docker-build: ## to build the docker image
-	@$(INFO) Performing Docker build...
+	@$(INFO) Performing Docker build ${APP_NAME}:${APP_VERSION}...
 	$(AT)$(DOCKER) build -f ${DOCKER_FILE} . \
 	-t ${APP_NAME}:${APP_VERSION} || ${FAIL}
 	@$(OK) Performing Docker build ${APP_NAME}:${APP_VERSION}
--- a/Makefile
+++ b/Makefile
@@ -41,6 +41,7 @@ GO                           := $(shell which go)
 GO_LDFLAGS                   += -X "github.com/mattermost/${APP_NAME}/service.buildHash=$(APP_COMMIT)"
 GO_LDFLAGS                   += -X "github.com/mattermost/${APP_NAME}/service.buildVersion=$(APP_VERSION)"
 GO_LDFLAGS                   += -X "github.com/mattermost/${APP_NAME}/service.buildDate=$(BUILD_DATE)"
+GO_LDFLAGS                   += -X "github.com/mattermost/${APP_NAME}/service.goVersion=$(GO_VERSION)"
 # Architectures to build for
 GO_BUILD_PLATFORMS           ?= linux-amd64 linux-arm64 darwin-amd64 darwin-arm64 freebsd-amd64
 GO_BUILD_PLATFORMS_ARTIFACTS = $(foreach cmd,$(addprefix go-build/,${APP_NAME}),$(addprefix $(cmd)-,$(GO_BUILD_PLATFORMS)))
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -51,12 +54,17 @@ GO_TEST_OPTS                 += -mod=readonly -failfast -race
...
+PROTECTED_BRANCH := main
+CURRENT_BRANCH   := $(shell git rev-parse --abbrev-ref HEAD)
 # Use repository name as application name
 APP_NAME    := $(shell basename -s .git `git config --get remote.origin.url`)
 # Get current commit
 APP_COMMIT  := $(shell git log --pretty=format:'%h' -n 1)
-# Check if we are in a release tag, if yes use it as app version.
-APP_VERSION := $(shell git describe --abbrev=0 --exact-match --tags > /dev/null 2>&1 || echo dev-$(APP_COMMIT))
+# Check if we are in a release tag, if yes use it as app version, else use dev-sha as app version
+APP_VERSION    := $(shell git describe --abbrev=0 --exact-match --tags > /dev/null 2>&1 || echo dev-$(APP_COMMIT))
```
pfltdv, line 65: Editorial review
```suggestion
APP_VERSION := $(shell git describe --abbrev=0 --exact-match --tags > /dev/null 2>&1 || echo dev-$(APP_COMMIT))
```

build/Dockerfile
```diff
@@ -1,7 +1,11 @@
 # This dockerfile is used to build Mattermost rtcd
-# A multi stage build, with golang:1.17.9-alpine used as a builder
+# A multi stage build, with golang:${GO_VERSION}-alpine used as a builder
 # and gcr.io/distroless/static as runner
-FROM golang:1.17.9-alpine@sha256:5c2fcfeb58ad9d4948d94e7b2d0432a9bc38bee0f8dfb41d383f38a18b75c38d as builder
+
+# GO_VERSION holds the go version we will build the image against
+# automatically generated from go.mod file using makefile
+ARG GO_VERSION=1.17
+FROM golang:${GO_VERSION}-alpine as builder
```
pfltdv, line 8: That is sweet and great idea.

pfltdv (approved):

Makefile
```diff
@@ -51,12 +54,17 @@ GO_TEST_OPTS                 += -mod=readonly -failfast -race
 GO_OUT_BIN_DIR               := ./dist
 
 ## General Variables
+# Branch Variables
+PROTECTED_BRANCH := main
```
streamer45, line 58: Should this be `master` or are we planning to switch? I am fine either way, in fact the plugin's repo defaults to `main`.

build/Dockerfile
```diff
@@ -1,7 +1,11 @@
 # This dockerfile is used to build Mattermost rtcd
-# A multi stage build, with golang:1.17.9-alpine used as a builder
+# A multi stage build, with golang:${GO_VERSION}-alpine used as a builder
 # and gcr.io/distroless/static as runner
-FROM golang:1.17.9-alpine@sha256:5c2fcfeb58ad9d4948d94e7b2d0432a9bc38bee0f8dfb41d383f38a18b75c38d as builder
+
+# GO_VERSION holds the go version we will build the image against
+# automatically generated from go.mod file using makefile
+ARG GO_VERSION=1.17
```
streamer45, line 7: What happens to the dot release in such case, is the latest one available used?

Makefile
```diff
@@ -108,15 +116,17 @@ help: ## to get help
 	awk 'BEGIN {FS = ":.*?## "}; {printf "make ${CYAN}%-30s${CNone} %s\n", $$1, $$2}'
 
 .PHONY: lint
-lint: go-lint go-mod-check docker-lint ## to lint all
```
streamer45, line 111: Could you explain why we are removing this?

streamer45 (commented):
Awesome, left a couple of comments :+1:

build/Dockerfile
spirosoik, line 8: we need to pin a version and SHA.

Makefile
phoinixgrr: Nice catch. I was under the impression that the protected branch was `main` for the rtcd project. Will change that to main 👍

Makefile
phoinixgrr: Will rebase and fix that! Thanks

Makefile
stafot: Ideally we should eventually move everywhere to the new naming conventions

build/Dockerfile
phoinixgrr: This will break all functionality with using dynamic values to go version, depending on the go.mod file. 
Thus will revert https://github.com/mattermost/rtcd/pull/36/commits/f9fe58f21820675babfa6f2b14b69541a0a2f98a

Makefile
phoinixgrr: `go-mod-check` is not linting, is it? 
If it is required , it can be triggered indepentanly with make go-mod-check. 

Or maybe create a new target named make checks/tests to include that process.

build/Dockerfile
phoinixgrr: Would you be ok with commiting this for now, and work on a future second iteration to providing a process to-pre populating SHA hashes for GOLANG images, and using those dynamically?

Makefile
phoinixgrr: Fixed!

Makefile
streamer45: Not linting but something we'd probably want to check as part of CI. So happy to move it out of here but would prefer to retain it.

phoinixgrr: Noted. Will create a seperate target for these kind of checks

stafot (dismissed):
LGTM

pfltdv (dismissed):
LGTM

build/Dockerfile
spirosoik: Yeah we discussed on Friday with Agniva and unfortunately we need always to pin to a specific version and SHA and after soak time of patch releases to adopt them.

phoinixgrr: Cool will drop the commit

phoinixgrr:
Dropped reviews because of dropping https://github.com/mattermost/rtcd/commit/f9fe58f21820675babfa6f2b14b69541a0a2f98a 

https://github.com/mattermost/rtcd/pull/36#discussion_r874453227

pfltdv (approved):

PR #38, opened 2022-05-17
[MM-44322] Pass configured ICEServers to peer connection

streamer45:
#### Summary

At some point we stopped passing ice servers (e.g. STUN urls) to the peer connection which meant server reflexive candidates were no longer generated. This is perfectly fine given we expect direct connectivity to be possible through the configured UDP port but if that's not the case there's still a chance things could work through the `srflx` candidate. 

This is likely what caused [this issue](https://github.com/mattermost/mattermost-plugin-calls/issues/73).

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-44322


cpoile (approved):
👍

PR #40, opened 2022-05-18
[fix] Makefile docker-sign

phoinixgrr:
#### Summary

This fixes the docker-sign target.

Cosign binary, was escaped and did not run inside the cosign dedicated container.
Moreover, we need the cosign key variable to be exposed in the container as a file.
Therefore we mount the required volumes

#### Ticket Link

Ticket: https://mattermost.atlassian.net/browse/DOPS-908

PR #41, opened 2022-05-18
[fix] Makefile docker-verify

phoinixgrr:
This fixes the `docker-verify` target which is affected by the same bug as described https://github.com/mattermost/rtcd/pull/40 
Cosign binary, was escaped and did not run inside the cosign dedicated container.

#### Ticket Link


pfltdv (approved):
LGTM

PR #42, opened 2022-05-19
[feat] Makefile go-test refactor

phoinixgrr:
#### Summary

- **Makefile:** Migrating `go-test` target, to running under a container, thus decoupling from local system
- **Makefile:** Dynamically extracting and using Go Version from go.mod file
- **Dockerfile**: Using `golang:1.17` instead of `golang:1.17-alpine` for `builder` stage. Thus removing the dependency of maintaining a list of packages. 
Dockerfile step: 
```
  RUN apk add \
    coreutils=9.0-r2 \
    make=4.3-r0 \
    git=2.34.2-r0 \
    --no-cache
```

#### Ticket Link

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,6 +22,7 @@ DOCKER_REGISTRY_REPO    += mattermost
 DOCKER_USER             ?= user
 DOCKER_PASSWORD         ?= password
 ## Docker Images
+DOCKER_IMAGE_GO         += "golang:${GO_VERSION}@sha256:79138c839452a2a9d767f0bba601bd5f63af4a1d8bb645bf6141bff8f4f33bb8"
 DOCKER_IMAGE_GOLINT     += "golangci/golangci-lint:v1.45.2@sha256:e84b639c061c8888be91939c78dae9b1525359954e405ab0d9868a46861bd21b"
 DOCKER_IMAGE_DOCKERLINT += "hadolint/hadolint:v2.9.2@sha256:d355bd7df747a0f124f3b5e7b21e9dafd0cb19732a276f901f0fdee243ec1f3b"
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -22,6 +22,7 @@ DOCKER_REGISTRY_REPO    += mattermost
 DOCKER_USER             ?= user
 DOCKER_PASSWORD         ?= password
 ## Docker Images
+DOCKER_IMAGE_GO         += "golang:${GO_VERSION}@sha256:79138c839452a2a9d767f0bba601bd5f63af4a1d8bb645bf6141bff8f4f33bb8"
```
pfltdv, line 25: Will same sha work for different go versions ?

phoinixgrr: It will not. 
The user will have to update the SHA hash, if the go version defined under `go.mod` file changes

I think it is good tradeoff between automating the go version extraction, and at the same time, pining the image using the SHA.

build/Dockerfile
```diff
@@ -1,20 +1,17 @@
...
 # Setup directories structure and compile
 COPY . /src
 WORKDIR /src
-RUN apk add \
-    coreutils=9.0-r2 \
-    make=4.3-r0 \
-    git=2.34.2-r0 \
-    --no-cache \
-    && make go-build
+RUN make go-build
```
pfltdv, line 14: :heart: Great catch.

phoinixgrr: Exactly, no need to use alpine there in the build, stage, since we keep only the binary compiled from the builder stage, and use distroless as runner

Makefile
pfltdv: Thank you very much, it is clear now. Yes we need that tradeoff at the moment.

pfltdv (approved):
:rocket: Great work!

PR #48, opened 2022-05-25
[fix] fix introduced on the Docker_Registry assigment

phoinixgrr:
#### Summary

fix introduced on the Docker_Registry assigment variable


pfltdv (approved):
LGTM

PR #53, opened 2022-05-27
[feat] Makefile release taging refactor

phoinixgrr:
#### Summary

We are changing the logic that the APP_VERSION variable is being generated.
Specifically:

- If we run under a protected_branch, we use `protected_branch_name-sha` as APP_VERSION
- If we run under a semver tag, we also push `latest` tag

Signed-off-by: Akis Maziotis <akis.maziotis@mattermost.com>

#### Ticket Link

Ticket: https://mattermost.atlassian.net/browse/DOPS-908

PR #54, opened 2022-05-27
[fix] Makefile make sure we also sign/verify latest

phoinixgrr:
#### Summary

Building on https://github.com/mattermost/rtcd/pull/53
We are also signing and verifying the `latest` tag.


pfltdv (approved):
:100: LGTM

PR #57, opened 2022-06-07
[MM-44897] Add draining wait on RTC server stop

streamer45:
#### Summary

We make stopping the RTC server a blocking operation so that we make sure to drain the instance and exit only when all call sessions are done. 

/cc @angeloskyratzakos 

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-44897


cpoile (approved):
👍

PR #60, opened 2022-06-09
[MM-44486] Use proper ICE server configs

streamer45:
#### Summary

Changing `ServerConfig.ICEServers` to be a list of proper ICE server config, potentially supporting TURN servers as well, which need credentials.

**Note**

This is a breaking change but it's one we can afford to make on this side. I've also added a backwards compatible unmarshaler for the config so that we can still support the previous format, just in case.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-44486

### DIFF
--- a/service/rtc/config.go
+++ b/service/rtc/config.go
@@ -5,6 +5,7 @@ package rtc
 
 import (
 	"fmt"
+	"strings"
 )
 
 type ServerConfig struct {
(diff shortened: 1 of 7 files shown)


service/rtc/config.go
```diff
@@ -55,3 +61,99 @@ func (c SessionConfig) IsValid() error {
...
+		}
+	}
+	return nil
+}
+
+func (s ICEServers) getSTUN() string {
+	for _, cfg := range s {
+		for _, u := range cfg.URLs {
+			if strings.HasPrefix(u, "stun:") {
+				return u
```
cpoile, line 102: I guess that's expected, to return only the first?

cpoile (approved):
Cool. I'm assuming there's no breaking changing to client config, so we don't have to worry about that. :)

service/rtc/config.go
streamer45: Yes, we just need one. We could randomize but not a big deal at the moment.

> Cool. I'm assuming there's no breaking changing to client config, so we don't have to worry about that. :)

On the client, I am still finishing up the PRs but we'll be adding a new ICEServerConfigs field along with the existing ICEServers which we'll deprecate. Will have PRs up tomorrow.

streamer45:
> Cool. I'm assuming there's no breaking changing to client config, so we don't have to worry about that. :)

On the client, I am still finishing up the PRs but we'll be adding a new ICEServerConfigs field along with the existing ICEServers which we'll deprecate. Will have PRs up tomorrow.

PR #62, opened 2022-06-16
[MM-45150] Fix failing to add tracks due to full channel

streamer45:
#### Summary

It took a while to debug this cause it wasn't clear at which level the problem was happening (client vs mattermost server vs plugin vs rtcd). Initial inspection of the logs at the various levels indicated that some clients failed to receive websocket events related to voice tracks which resulted in them not hearing some unmuted participants. In fact, when someone unmutes we loop through all connections to broadcast the necessary events that eventually get the voice track to the client.

There were around 130 participants in COM yesterday. One of those unmuting caused 130 track events to get sent down a channel of buffer 64 which wasn't able to handle them all in time and got full. Eventually messages for about 4/5 users were dropped. 

Finding an ideal value is a hard problem of course and we certainly don't want to make the operation blocking but it looks like that, to be on the safe side, the buffer size should roughly be equal to the maximum burst of events we expect which roughly translates into the maximum number of participants in a call. Of course it's a little more complex than this since multiple calls can live on a single rtcd instance but unmuting events for separate calls are unlikely to happen simultaneously. I think we are bumping them sufficiently high here but will keep monitoring in case they needs further tuning.

Issues addressed in this PR:

- Making client error channels buffered.
- Logging out the original error in case the error channel gets full so we don't lose possibly valuable information.
- Setting the client receive channel size to match the websocket one since they are tightly coupled. By client here we mean the rtcd client on the plugin side which potentially could handle multiple calls and hundreds of connected users.

Once this is merged I'll tag it and create one on the plugin side since these are client related changes so will need to a new plugin release.

Special thanks to @agnivade for providing client-side logs that helped me debugging this. 

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-45150

### DIFF
--- a/service/client.go
+++ b/service/client.go
@@ -267,6 +267,12 @@ func (c *Client) Close() error {
 }
 
 func (c *Client) sendError(err error) {
+	c.mut.RLock()
+	defer c.mut.RUnlock()
+	if c.closed {
+		return
+	}
+
 	select {
 	case c.errorCh <- err:
 	default:
--- a/service/client.go
+++ b/service/client.go
@@ -18,7 +18,6 @@ import (
 )
 
 const (
-	msgChSize                = 64
 	maxReconnectInterval     = 30 * time.Second
 	defaultReconnectInterval = 2 * time.Second
 )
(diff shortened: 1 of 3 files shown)


cpoile (approved):
Nice debugging work, must have been tricky. 😅

phoinixgrr:
Awesome description!!!! Thanks 🚀

PR #63, opened 2022-06-17
[MM-45001] Add support for generating short-lived TURN credentials

streamer45:
#### Summary

Adding support for generating short-lived TURN credentials as described in https://datatracker.ietf.org/doc/html/draft-uberti-behave-turn-rest-00#section-2.2

#### Related PRs

https://github.com/mattermost/mattermost-plugin-calls/pull/116

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-45001

### DIFF
--- a/service/rtc/config.go
+++ b/service/rtc/config.go
@@ -17,6 +17,11 @@ type ServerConfig struct {
 	ICEHostOverride string `toml:"ice_host_override"`
 	// A list of ICE server (STUN/TURN) configurations to use.
 	ICEServers ICEServers `toml:"ice_servers"`
+	// The secret key used to generate TURN short-lived authentication
+	// credentials.
+	TURNStaticAuthSecret string
+	// The number of minutes that the generated TURN credentials will be valid for.
+	TURNCredentialsExpirationMinutes int
 }
 
 func (c ServerConfig) IsValid() error {
@@ -26,6 +31,10 @@ func (c ServerConfig) IsValid() error {
 
 	if err := c.ICEServers.IsValid(); err != nil {
 		return fmt.Errorf("invalid ICEServers value: %w", err)
+	}
+
+	if c.TURNStaticAuthSecret != "" && c.TURNCredentialsExpirationMinutes <= 0 {
+		return fmt.Errorf("invalid TURNCredentialsExpirationMinutes value: should be a positive number")
 	}
 
 	return nil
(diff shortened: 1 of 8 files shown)


service/rtc/config.go
```diff
@@ -79,13 +88,31 @@ func (c ICEServerConfig) IsValid() error {
...
 			return fmt.Errorf("URL is not a valid STUN/TURN server")
 		}
 	}
 	return nil
 }
 
+func (c ICEServerConfig) IsTURN() bool {
+	for _, u := range c.URLs {
+		if strings.HasPrefix(u, "turn:") {
+			return true
```
cpoile, line 101: break early, so only one needs to be valid? (same Q for `IsSTUN` below)

service/rtc/config.go
```diff
@@ -97,10 +124,8 @@ func (s ICEServers) IsValid() error {
 
 func (s ICEServers) getSTUN() string {
 	for _, cfg := range s {
-		for _, u := range cfg.URLs {
-			if strings.HasPrefix(u, "stun:") {
-				return u
-			}
+		if cfg.IsSTUN() && len(cfg.URLs) > 0 {
+			return cfg.URLs[0]
```
cpoile, line 128: This might mean: if there's an incorrect first url but a correct second url, then we just returned the incorrect one?

cpoile (commented):
Left a comment, maybe a bug?

service/rtc/config.go
streamer45: I was assuming correctness there really considering it would fail eventually when creating the connection but makes sense to fail as soon as possible especially since these utils are used in `IsValid` as well. Will send a fix, thanks!

cpoile (approved):
Thanks, looks great!

service/rtc/turn.go
```diff
@@ -0,0 +1,57 @@
...
+	if username == "" {
+		return "", "", fmt.Errorf("username should not be empty")
+	}
+
+	if secret == "" {
+		return "", "", fmt.Errorf("secret should not be empty")
+	}
+
+	if expirationTS <= 0 {
+		return "", "", fmt.Errorf("expirationTS should be a positive number")
```
oh6hay, line 24: should a maximum age be imposed?

oh6hay (approved):
LGTM

service/rtc/turn.go
streamer45: It won't hurt I guess but finding a value may be a bit of a guesswork. Given the recommended value is 1 day we could put a week (7 days) maybe? I can't think of any use-case to go higher than that to be honest.

cpoile (approved):

PR #65, opened 2022-07-21
[MM-43838] WebSocket reconnection support

streamer45:
#### Summary

On this side it's pretty simple as it boils down to updating the connections map so that further messages will be sent to the right node.

Unrelated, but caught during testing, this also fixes a potential leak in case of initialing signaling timeout.

#### Related PRs

https://github.com/mattermost/mattermost-plugin-calls/pull/145

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-43838

### DIFF
--- a/service/service.go
+++ b/service/service.go
@@ -290,6 +290,22 @@ func (s *Service) handleClientMsg(msg ws.Message) error {
 		s.mut.Unlock()
 
 		return nil
+	case ClientMessageReconnect:
+		data, ok := cm.Data.(map[string]string)
+		if !ok {
+			return fmt.Errorf("unexpected data type: %T", cm.Data)
+		}
+		sessionID := data["sessionID"]
+		if sessionID == "" {
+			return fmt.Errorf("missing sessionID in client message")
+		}
+
+		s.log.Debug("reconnect message, updating connMap", mlog.String("sessionID", sessionID))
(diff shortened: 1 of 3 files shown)


service/service.go
```diff
@@ -289,6 +289,22 @@ func (s *Service) handleClientMsg(msg ws.Message) error {
...
+			return fmt.Errorf("unexpected data type: %T", cm.Data)
+		}
+		sessionID := data["sessionID"]
+		if sessionID == "" {
+			return fmt.Errorf("missing sessionID in client message")
+		}
+
+		s.log.Debug("reconnect message, updating connMap", mlog.String("sessionID", sessionID))
+		s.mut.Lock()
+		s.connMap[sessionID] = msg.ConnID
```
cpoile, line 305: I know we're doing it above too, but is there any chance we won't have a ConnID, and should check for it?

cpoile (approved):
Nice!

service/service.go
streamer45: The `ConnID` comes straight from the websocket server so it must be set.
Are you asking about the case in which a reconnect occurs but no previous session is in the map?

cpoile: The first -- I haven't looked at the server websocket code. :)

Cool, thanks!

PR #69, opened 2022-08-04
service/rtc: Fixing repeat initialization of connection

agnivade:
MultiConn was being incorrectly initialized inside a loop.
This created repeated listener goroutines of the same connection.

### DIFF
--- a/service/rtc/server.go
+++ b/service/rtc/server.go
@@ -154,10 +154,11 @@ func (s *Server) Start() error {
 		}
 
 		conns = append(conns, udpConn)
-		s.udpConn, err = newMultiConn(conns)
-		if err != nil {
-			return fmt.Errorf("failed to create multiconn: %w", err)
-		}
+	}
+	var err error
+	s.udpConn, err = newMultiConn(conns)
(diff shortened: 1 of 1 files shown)


streamer45 (dismissed):
Thanks for spotting this :raised_hands:  I'll run a load-test before merging just to make extra sure all is good since unfortunately our initial perf tests were conducted with the bogus implementation although I don't expect to see substantial differences.

streamer45:
Load-tests look good. The bug could have easily had a much worse impact especially since the "leaked" readers would still receive data from the underlying connection but never deliver it. Fortunately they would block on channel send so at most one datagram would be lost per reader which is likely the reason this passed unnoticed for so long.

cpoile (approved):
Thanks @agnivade!

agnivade:
@streamer45 - It seems like it requires commits to be signed to merge a PR. Do you want me to squash and force-push?

agnivade:
@cpoile - can you please approve once more? Thanks!

cpoile (approved):
👍

PR #70, opened 2022-08-17
Getting started guide

streamer45:
#### Summary

Wrote a quick guide on how to get started with Calls + `rtcd`.

@cpoile Would you be able to give it a quick try on your end as I fear there may be something to adapt in case of Mac users as the docker IPs may be different, not sure.

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -1,0 +1,62 @@
+# Running calls through `rtcd`
+
+## Prerequisites
+
+- An [Enterprise](https://docs.mattermost.com/about/editions-and-offerings.html#mattermost-enterprise) licensed Mattermost installation with [Calls](https://github.com/mattermost/mattermost-plugin-calls) installed.
+- A working version of [Docker](https://www.docker.com/) installed.
+
+## Installation
+
+As first step we fetch the latest official `rtcd` docker image:
+
+```sh
+docker pull mattermost/rtcd:latest
+```
+
+## Running
+
+To start `rtcd` we can run the following command:
+
+```sh
+docker run --name rtcd -e "RTCD_LOGGER_ENABLEFILE=false" -e "RTCD_API_SECURITY_ALLOWSELFREGISTRATION=true" -p 8443:8443/udp -p 8045:8045/tcp mattermost/rtcd
+```
+
+**Note**
+
+- By default the service starts even if no configuration file is provided. In such case default values are used. In the example above we are overriding a couple of config settings:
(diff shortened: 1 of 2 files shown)


docs/getting_started.md
```diff
@@ -0,0 +1,62 @@
+# Running calls through `rtcd`
+
+## Prerequisites
+
+- An [Enterprise](https://docs.mattermost.com/about/editions-and-offerings.html#mattermost-enterprise) licensed Mattermost installation with [Calls](https://github.com/mattermost/mattermost-plugin-calls) installed.
+- A working version of [Docker](https://www.docker.com/) installed.
+
+## Installation
```
cpoile, line 8: Can we have two versions of this -- one with docker (as you have) and an alternative "No docker" version? I don't use docker (it's horrible on m1 macs. I had to install it for this, but I'll delete it after). So I just use `make go-run` and it works.

docs/getting_started.md
```diff
@@ -0,0 +1,62 @@
...
+```sh
+docker run --name rtcd -v /path/to/rtcd/config:/config mattermost/rtcd -config /config/config.toml
+```
+
+### Verify service is running
+
+Finally, to verify that the service is correctly running we can try calling the HTTP API:
+
+```sh
+curl http://172.17.0.1:8045/version
```
cpoile, line 46: for me it was (and probably will be for others too, right?):
```
curl http://localhost:8045/version
```

docs/getting_started.md
```diff
@@ -0,0 +1,62 @@
...
+
+This should return a JSON object with basic information about the service such as its build version.
+
+## Configuration
+
+Configuration for the service is fully documented in-place through the [`config.sample.toml`](config/config.sample.toml) file.
+
+## Running calls
+
+The last step to get calls working through `rtcd` is to configure the Calls side to use the service. This is done via the **Admin Console -> Plugins -> Calls -> RTCD service URL** setting, which in this example will be set to `http://testClient:@172.17.0.1:8045`.
```
cpoile, line 57: like above, my address is `localhost`

cpoile (approved):
Some possible changes, but otherwise looks good!

docs/getting_started.md
streamer45: Oh well, since we are exposing the port to the host it probably works just fine. Will update it, thanks.

docs/getting_started.md
streamer45: We totally can although this guide is really targeting users that want to deploy the service rather than developing it locally.

PR #73, opened 2022-09-08
[MM-46918] Update docs

streamer45:
#### Summary

Updating the docs as per https://github.com/mattermost/mattermost-plugin-calls/pull/186

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-46918

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -54,9 +54,9 @@ Configuration for the service is fully documented in-place through the [`config.
 
 ## Running calls
 
-The last step to get calls working through `rtcd` is to configure the Calls side to use the service. This is done via the **Admin Console -> Plugins -> Calls -> RTCD service URL** setting, which in this example will be set to `http://testClient:@localhost:8045`.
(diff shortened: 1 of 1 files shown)


cpoile (approved):
Oh, I just realized (re: the PR this is documenting) -- I've had to change the client id because for some reason rtcd didn't let me reuse one that I had used before. Is that a possible problem, or did I just make a mistake somewhere?

streamer45:
> Oh, I just realized (re: the PR this is documenting) -- I've had to change the client id because for some reason rtcd didn't let me reuse one that I had used before. Is that a possible problem, or did I just make a mistake somewhere?

@cpoile Sounds like a conflict but would need logs to confirm. By default `rtcd` stores credentials through a file so unless that's deleted they are persisted indefinitely. On the plugin side credentials are stored in the k/v store. If the entry gets deleted (e.g. through a db reset) the same client id cannot be re-used again until a similar reset happens on the `rtcd` side. 

There are other ways to get locked out but slightly more tricky to trigger (e.g. what happened in Cloud the other week). A change in URL would result in an attempt to re-register and if this succeeds then new credentials are saved, replacing the previous ones. But realistically this requires separate `rtcd` instances to happen.

Please let me know if it happens again so we can debug what's going on.

PR #74, opened 2022-09-09
Use a sync.Pool to reuse read buffers

agnivade:


agnivade:
@streamer45 - would you be able to run some perf tests on this?

service/rtc/sfu.go
```diff
@@ -254,9 +259,17 @@ func (s *Server) InitSession(cfg SessionConfig, closeCb func() error) error {
...
-					s.log.Error("failed to read RTP packet", mlog.Err(readErr))
+				buf := s.bufPool.Get().([]byte)
+				i, _, err := remoteTrack.Read(buf)
+				if err != nil {
+					s.log.Error("failed to read RTP packet", mlog.Err(err))
+					s.metrics.IncRTCErrors(us.cfg.GroupID, "rtp")
+					return
+				}
+
+				rtp := &rtp.Packet{}
```
agnivade, line 270: wait, I need to check if this escapes to heap. And if there's a way to make it non-escaping. But that shouldn't block the load tests.

agnivade: Does not escape. All good. :+1:

streamer45 (approved):
Looks good :+1: 

Had a couple of load-tests which didn't show any significant change which is expected I feel given the low level nature of this change but conceptually we are allocating less times so it should improve overall performance as we scale further.

PR #75, opened 2022-09-30
[MM-44220] Implement client session tokens

m-ripper:
#### Summary

Currently, basic authentication is done with every request, which means that time-consuming processes like the comparison of hashes are done on a request basis. In order to reduce the load on the server side and improve performance, it makes sense to implement authentication with temporary session tokens.

Bearer authentication was implemented which allows skipping time-consuming processes like the hash-comparison.
Clients can now log in using the POST /login endpoint by providing their basic credentials as JSON body and receive a temporary session token in return. This token can be used to authenticate the client for all other endpoints of this service.

#### Ticket Link

Related to https://mattermost.atlassian.net/browse/MM-44220

#### Release Note

```release-note
Added a new config setting: `security.session_cache.expiration_minutes` that determines how long a session is valid until it expires. The default value is 1 day.

Added new API endpoint POST /login that expects credentials as JSON in the body.

Added bearer-authentication: clients can login via the POST /login endpoint and use the resulting token in further requests to authenticate.
```


streamer45 (changes requested):
Thanks @m-ripper , code looks good overall :+1: I particularly liked how you tried to keep everything consistent with the existing code and styling format. Adding tests and documentation was another great idea :100: 

Other than the minor comments that I left inline there are a couple of things we should address before merging:

1. We should make the `sessionCache` goroutine safe. The API (and the service as a whole) is meant to receive multiple concurrent requests so access to any shared resource (the session map in this case) should be guarded accordingly.
2. We should handle the `unregister` case so that we invalidate the token(s) for the client, otherwise they can keep using it after the account being removed which doesn't make much sense.

service/auth/session_cache.go
```diff
@@ -0,0 +1,89 @@
...
+		return nil, fmt.Errorf("token is invalid")
+	}
+	if time.Now().After(session.ExpirationDate) {
+		delete(t.sessionMap, token)
+		return nil, fmt.Errorf("session is expired")
+	}
+	return &session, nil
+}
+
+func (t *SessionCache) Put(clientID string, authKey string) (string, error) {
```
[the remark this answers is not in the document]
m-ripper: Oh, that was probably some leftover that I forgot to remove, I will get rid of it, thanks!

service/auth/session_cache.go
```diff
@@ -0,0 +1,89 @@
...
+
+func (t *SessionCache) newUniqueToken() (string, error) {
+	token, err := newRandomString(MinKeyLen)
+	if err != nil {
+		return "", err
+	}
+	// If the bearer token already exists, try again
+	if _, ok := t.sessionMap[token]; ok {
+		return t.newUniqueToken()
+	}
```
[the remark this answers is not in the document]
m-ripper: The chances for a conflict are very low, but in theory, it could happen, what do you think, is it worth counting it as a risk?

service/auth/session_cache.go
```diff
@@ -0,0 +1,98 @@
...
```
streamer45, line 96: We need to guard this access as well.

service/auth.go
```diff
@@ -48,6 +58,38 @@ func (s *Service) authHandler(w http.ResponseWriter, r *http.Request) (clientID
...
+}
+
+func bearerAuth(r *http.Request) (token string, ok bool) {
+	auth := r.Header.Get("Authorization")
+	return parseBearerAuth(auth)
+}
+
+func parseBearerAuth(auth string) (token string, ok bool) {
+	// Case-insensitive prefix match
+	if len(auth) < len(bearerPrefix) || !strings.EqualFold(strings.ToLower(auth[:len(bearerPrefix)]), strings.ToLower(bearerPrefix)) {
```
streamer45, line 87: Why do we lowercase the strings? I'd expect `EqualFold` to take care of that already.

service/auth/session_cache.go
```diff
@@ -0,0 +1,102 @@
...
+	t.mut.Unlock()
+	return token, nil
+}
+
+func (t *SessionCache) Delete(clientID string) {
+	t.mut.RLock()
+	defer t.mut.RUnlock()
+	for token, session := range t.sessionMap {
+		if session.ClientID == clientID {
+			t.mut.Lock()
```
streamer45, line 82: I'd expect this to deadlock (locking on read lock). We should unlock prior to locking again. Or simply lock once and accept it being locked for the whole duration of the function.

service/auth.go
```diff
@@ -27,6 +30,13 @@ func (s *Service) authHandler(w http.ResponseWriter, r *http.Request) (clientID
 		s.httpAudit("authHandler", data, nil, r)
 	}()
 
+	if strings.HasPrefix(r.Header.Get("Authorization"), bearerPrefix) {
+		return s.bearerAuthHandler(w, r)
+	}
+	return s.basicAuthHandler(w, r)
+}
+
+func (s *Service) basicAuthHandler(w http.ResponseWriter, r *http.Request) (clientID string, code int, err error) {
```
streamer45, line 39: I think we can safely remove named return values here as they are not used.

service/auth.go
```diff
@@ -48,6 +58,37 @@ func (s *Service) authHandler(w http.ResponseWriter, r *http.Request) (clientID
 	return clientID, http.StatusOK, nil
 }
 
+func (s *Service) bearerAuthHandler(w http.ResponseWriter, r *http.Request) (clientID string, code int, err error) {
```
streamer45, line 61: Same thing here.

service/auth.go
```diff
@@ -48,6 +58,37 @@ func (s *Service) authHandler(w http.ResponseWriter, r *http.Request) (clientID
...
+		return "", http.StatusUnauthorized, fmt.Errorf("authentication failed: unauthorized")
+	}
+
+	return clientID, http.StatusOK, nil
+}
+
+func bearerAuth(r *http.Request) (token string, ok bool) {
+	auth := r.Header.Get("Authorization")
+	return parseBearerAuth(auth)
+}
```
streamer45, line 83: Do we really need this function? It's just about passing `r.Header.Get("Authorization")` to `parseBearerAuth()` which can probably be done directly from the caller.

service/auth.go
```diff
@@ -128,6 +170,37 @@ func (s *Service) unregisterClient(w http.ResponseWriter, r *http.Request) {
...
+	clientID := data.reqData["clientID"]
+	authKey := data.reqData["authKey"]
+	bearerToken, err := s.auth.Login(clientID, authKey)
+	if err != nil {
+		data.err = err.Error()
+		data.code = http.StatusBadRequest
+		return
+	}
 
+	s.log.Debug("logged in client", mlog.String("clientID", clientID), mlog.String("bearerToken", bearerToken))
```
streamer45, line 203: We probably don't want to log authentication tokens. The client ID is sufficient.

service/auth/session_cache.go
```diff
@@ -0,0 +1,100 @@
...
+func (t *SessionCache) newUniqueToken() (string, error) {
+	token, err := newRandomString(MinKeyLen)
+	if err != nil {
+		return "", err
+	}
+	t.mut.RLock()
+	defer t.mut.RUnlock()
+	// If the bearer token already exists, try again
+	if _, ok := t.sessionMap[token]; ok {
+		return t.newUniqueToken()
```
streamer45, line 97: This looks like it could cause potential recursive locking which is likely not what we want to do here. I'd probably suggest to avoid the `defer` in this case and simply lock/unlock as needed.

service/auth/session_cache.go
```diff
@@ -0,0 +1,100 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	// Make sure there is only one bearer token per client.
+	t.Delete(clientID)
+
+	t.mut.Lock()
+	t.sessionMap[token] = CachedSession{ClientID: clientID, ExpirationDate: time.Now().Add(time.Duration(t.cfg.ExpirationMinutes) * time.Minute)}
+	t.mut.Unlock()
```
streamer45, line 73: Just noticing that we lock/unlock to delete just to lock/unlock again to set the session. Could you think of a simple solution to prevent this?

streamer45 (commented):
Thanks! Just gave this another pass, nothing major but a couple of possible improvements.

service/auth/session_cache.go
```diff
@@ -0,0 +1,104 @@
...
+
+func (c SessionCacheConfig) IsValid() error {
+	if c.ExpirationMinutes <= 0 {
+		return fmt.Errorf("invalid ExpirationMinutes value: should be a positive number")
+	}
+	return nil
+}
+
+type SessionCache struct {
+	cfg        *SessionCacheConfig
```
streamer45, line 29: I'd suggest using a value instead of a pointer.

streamer45 (approved):
Thanks for addressing everything @m-ripper, looks great :+1:

streamer45 (dismissed):

service/auth/session_cache.go
```diff
@@ -0,0 +1,104 @@
...
+	if time.Now().After(session.ExpirationDate) {
+		t.mut.Lock()
+		delete(t.sessionMap, token)
+		t.mut.Unlock()
+		return CachedSession{}, fmt.Errorf("session is expired")
+	}
+	return session, nil
+}
+
+func (t *SessionCache) Put(clientID string) (string, error) {
```
cpoile, line 58: Do you think `Put` has multiple responsibilities here?
Specifically, do you think there's a way to split this up (or simplify it) so that the calling code has to be more specific (and descriptive) about what it's actually intending to do?

service/auth/session_cache.go
```diff
@@ -0,0 +1,104 @@
...
+
+	token, err := t.newUniqueToken()
+	if err != nil {
+		return "", err
+	}
+
+	t.mut.Lock()
+	// Make sure there is only one bearer token per client.
+	t.delete(clientID)
+	t.sessionMap[token] = CachedSession{ClientID: clientID, ExpirationDate: time.Now().Add(time.Duration(t.cfg.ExpirationMinutes) * time.Minute)}
```
cpoile, line 71: nit:
```suggestion
	t.sessionMap[token] = CachedSession{
		ClientID: clientID,
		ExpirationDate: time.Now().Add(time.Duration(t.cfg.ExpirationMinutes) * time.Minute),
	}
```

cpoile (commented):
Nice work Max! Nice tests, and the style matches what we're doing, thank you.
I think Claudio covered most of the major things. I have left one comment, and a couple really minor nits. 
Let me know how you want to deal with them (or not) and I'll quickly review again.

@streamer45 Before enabling this, I think we should think about a way to ensure that the clients are either 1) on an internal network, or 2) the connection is tls.  (re: [bearer tokens are a terrible idea](https://web.archive.org/web/20160329174534/http://hueniverse.com/2010/09/29/oauth-bearer-tokens-are-a-terrible-idea/)
Otherwise this is very insecure. Not sure if that's a concern or not? (or even how we would test/block on 1)

Second, if I'm reading the code correctly, we're not actually authenticating anything other than the client registration/unregistration and the ws upgrade, correct? We should be authenticating the rest of the endpoints, right?

service/auth/session_cache.go
m-ripper: First of all, I agree that there are multiple responsibilities, the generation of the token and the actual update of the cached session. I thought about splitting this up but I am not sure about it, what do you think?
Another solution would be to rename the method to `Cache` but this is probably not more descriptive.

service/auth/service.go
```diff
@@ -71,5 +76,22 @@ func (s *Service) Unregister(id string) error {
...
+	s.sessionCache.Delete(id)
+
 	return nil
 }
+
+func (s *Service) Login(id, key string) (string, error) {
+	if err := s.Authenticate(id, key); err != nil {
+		return "", fmt.Errorf("login failed: %w", err)
+	}
+	bearerToken, err := s.sessionCache.NewUniqueToken()
```
streamer45, line 89: If we are using it like this I am thinking we may as well decouple it completely from the cache as it becomes more of a utility function that can safely live in the `auth` package. What do you think?

m-ripper: That could be done but it's not guaranteed anymore that the new token is unique then.

cpoile (commented):
Almost there, thanks Max for the great revisions!

streamer45 (dismissed):

cpoile (approved):
Awesome work, thanks @m-ripper!

streamer45 (approved):

streamer45:
@m-ripper Unfortunately this repository is rather strict when it comes to merging. So it requires all commits to be signed. This likely means you need to recommit everything signed (maybe squashed to save time) and force push.

m-ripper:
> @m-ripper Unfortunately this repository is rather strict when it comes to merging. So it requires all commits to be signed. This likely means you need to recommit everything signed (maybe squashed to save time) and force push.

That's no problem, I will take care of that as soon as I've setup signed commits. Thanks for clarifying!

streamer45 (approved):

PR #76, opened 2022-10-26
MM-47231 - Secure the /unregister api path

cpoile:
#### Summary
(just cutting and pasting the description of the problem from the channel 😄  ) :
- If client [self registration](https://github.com/mattermost/rtcd/blob/dded04f0cc2daee07ba58104c87f1422e493a0f9/config/config.sample.toml#L13) is enabled then a client can unregister itself. Right now an authenticated client can unregister any other client just by knowing their id.
- If self registration is disabled and an [admin key](https://github.com/mattermost/rtcd/blob/dded04f0cc2daee07ba58104c87f1422e493a0f9/config/config.sample.toml#L19-L21) is configured then that can unregister anyone.
- If none of the above is set (self register disabled and no admin configured) then the call should simply return forbidden.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-47231

### DIFF
--- a/service/auth.go
+++ b/service/auth.go
@@ -100,7 +100,17 @@ func (s *Service) unregisterClient(w http.ResponseWriter, r *http.Request) {
 	}
 	defer s.httpAudit("unregisterClient", data, w, r)
 
-	_, code, err := s.authHandler(w, r)
+	// If an admin client is not enabled, and self registration is not allowed,
+	// clients cannot unregister themselves.
+	if !s.cfg.API.Security.EnableAdmin && !s.cfg.API.Security.AllowSelfRegistration {
+		s.log.Warn("/unregister was called, but enable_admin and allow_self_registration are both false")
+		data.err = "unregister not enabled"
+		data.code = http.StatusForbidden
+		return
+	}
+
+	// Check if admin authKey or clientID + authKey have been provided
+	authedClientID, code, err := s.authHandler(w, r)
 	if err != nil {
 		data.err = err.Error()
 		data.code = code
@@ -117,6 +127,14 @@ func (s *Service) unregisterClient(w http.ResponseWriter, r *http.Request) {
 	if clientID == "" {
 		data.err = "client id should not be empty"
 		data.code = http.StatusBadRequest
+		return
+	}
+
+	// an authedClientID == "" means admin. So if there is an authedClientID,
+	// then the requested clientID needs to be the same.
+	if authedClientID != "" && authedClientID != clientID {
+		data.err = "client id not valid"
+		data.code = http.StatusForbidden
 		return
 	}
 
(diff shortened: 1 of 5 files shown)


service/auth.go
```diff
@@ -100,7 +100,16 @@ func (s *Service) unregisterClient(w http.ResponseWriter, r *http.Request) {
 	}
 	defer s.httpAudit("unregisterClient", data, w, r)
 
-	_, code, err := s.authHandler(w, r)
+	// If an admin client is not enabled, and self registration is not allowed,
+	// clients cannot unregister themselves.
+	if !s.cfg.API.Security.EnableAdmin && !s.cfg.API.Security.AllowSelfRegistration {
+		data.err = "enable_admin and allow_self_registration are both false"
```
cpoile, line 106: Not sure how much feedback you want to give in APIs -- I always err on the side of more info for API consumers (as long as it's not security-sensitive).

service/helper_test.go
```diff
@@ -78,6 +59,33 @@ func SetupTestHelper(tb testing.TB) *TestHelper {
 	return th
 }
 
+func MakeDefaultCfg(tb testing.TB) *Config {
```
streamer45, line 62: Should we call `tb.Helper()` in here?

service/auth.go
streamer45: This is not an user facing API as the client in this case is the plugin so I am not even sure how much of this would be surfaced but generally I think it would be best to log this information rather than returning it as this is not an actionable error from the client's perspective.

service/auth.go
```diff
@@ -120,6 +129,14 @@ func (s *Service) unregisterClient(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// an authedClientID == "" means admin. So if there is an authedClientID,
+	// then the requested clientID needs to be the same.
+	if authedClientID != "" && authedClientID != clientID {
+		data.err = "client id not valid"
+		data.code = http.StatusBadRequest
```
streamer45, line 136: I think this case would fit better in the 403 code since there's nothing wrong with the request itself but the server simply cannot go ahead due to missing authorization. Not a big deal though.

streamer45 (dismissed):
Thanks @cpoile , looks good. And great work on including thorough test cases :muscle:

service/auth.go
cpoile: Good point, I'll `log.Warn` the issue.

service/auth.go
cpoile: Yep, makes sense, thanks!

streamer45 (approved):

streamer45 (approved):

streamer45 (approved):

PR #81, opened 2022-12-16
[MM-47006] Add support for call quality measurements

streamer45:
#### Summary

Not as much happening on this side but a key change is that we need to read RTCP packets for both senders/receivers in order to generate reports through the configured interceptors.

PR also removes some spurious errors that would happen every time a disconnection happened.

#### Related PR

https://github.com/mattermost/mattermost-plugin-calls/pull/283

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-47006


cpoile (dismissed):
needs more bufpool

cpoile (commented):
This is weird, I'm not seeing any rtcp-related code in the changes. I'm only seeing +21 -12, all presumably unrelated code?

cpoile (approved):
Ah, ok. 👍

PR #82, opened 2022-12-19
Fix potential data races

streamer45:
#### Summary

PR fixes a couple of potential data races:

1. Cloning HTTP requests headers to avoid reading (logging) and writing at the same time.
2. Although the underlying `bitcask` store is already safe for concurrent use, the way we use it is not since we force a sync on every write.

### DIFF
--- a/service/audit.go
+++ b/service/audit.go
@@ -42,12 +42,13 @@ func (s *Service) httpAudit(handler string, data *httpData, w http.ResponseWrite
 }
 
 func reqAuditFields(req *http.Request) []mlog.Field {
-	delete(req.Header, "Authorization")
+	hdr := req.Header.Clone()
+	delete(hdr, "Authorization")
 	fields := []mlog.Field{
 		mlog.String("remoteAddr", req.RemoteAddr),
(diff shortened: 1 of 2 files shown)


cpoile (approved):
Cool. I guess it shouldn't matter that we're using it in a bit of a weird way (syncing on everything) because it's not a heavy r/w load. But maybe we should put a comment near the top, like a "warning: this is intended to be used for very light workloads..."?

streamer45:
> Cool. I guess it shouldn't matter that we're using it in a bit of a weird way (syncing on everything) because it's not a heavy r/w load. But maybe we should put a comment near the top, like a "warning: this is intended to be used for very light workloads..."?

We are consciously favouring consistency (preventing data loss) over performance. This is even [allowed by the framework](https://pkg.go.dev/github.com/prologic/bitcask#WithSync) but somehow it doesn't apply to deletes hence the use of sync at this level. 

But performance is a relative concept. We could easily set thousands of records per second with the existing approach which is several orders of magnitude more than what the service requires.

PR #83, opened 2022-12-20
[MM-49236] Implement server side VAD

streamer45:
#### Summary

It's been a while since I wanted to try this. PR implements server side voice activity detection. This is achieved by reading audio levels coming from clients through the special [`ssrc-audio-level`](https://www.rfc-editor.org/rfc/rfc6464) RTP header extension.

The main benefits of this approach are:

1. It avoids the performance hit needed by the client-side VAD implementation.
2. It should make mobile VAD work without additional changes.
3. The logic feels a bit more robust now. I am using standard deviation of sampled audio levels to figure out voice vs background noise. It may need some further tuning as I couldn't test it on too many systems but it's working decently well so far.

#### Related PR

https://github.com/mattermost/mattermost-plugin-calls/pull/285

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-49236

### DIFF
--- a/service/rtc/sfu.go
+++ b/service/rtc/sfu.go
@@ -44,6 +44,7 @@ var (
 
 const (
 	nackResponderBufferSize = 256
+	audioLevelExtensionURI  = "urn:ietf:params:rtp-hdrext:ssrc-audio-level"
 )
 
 func initMediaEngine() (*webrtc.MediaEngine, error) {
@@ -127,6 +128,12 @@ func (s *Server) InitSession(cfg SessionConfig, closeCb func() error) error {
 	i, err := initInterceptors(m)
 	if err != nil {
 		return fmt.Errorf("failed to init interceptors: %w", err)
+	}
+
+	if err := m.RegisterHeaderExtension(webrtc.RTPHeaderExtensionCapability{
+		URI: audioLevelExtensionURI,
+	}, webrtc.RTPCodecTypeAudio); err != nil {
+		return fmt.Errorf("failed to register header extension: %w", err)
 	}
 
 	sEngine := webrtc.SettingEngine{}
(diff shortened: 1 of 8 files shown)


service/rtc/sfu.go
```diff
@@ -275,16 +297,26 @@ func (s *Server) InitSession(cfg SessionConfig, closeCb func() error) error {
...
 
+				if us.vadMonitor != nil && audioLevelExtensionID > 0 {
+					var ext rtp.AudioLevelExtension
+					audioExtData := packet.GetExtension(uint8(audioLevelExtensionID))
+					if err := ext.Unmarshal(audioExtData); err != nil {
+						s.log.Error("failed to unmarshal audio level extension",
+							mlog.Err(err), mlog.String("sessionID", us.cfg.SessionID))
+					}
+					us.vadMonitor.PushAudioLevel(ext.Level)
+				}
```
cpoile, line 316: This is really cool -- but what happens when you mute while in a voiceOn state? 

One possible solution? Maybe we need send a voiceOff on mute. It may be redundant sometimes, but I think the clients will gracefully handle it.

cpoile (commented):
This is probably the coolest code you've done in at least a few days. :) Nice work! 

I think I found one bug?

I tested with mobile, and it works well, thank you!

service/rtc/sfu.go
streamer45: Thanks. For some reason I thought we were being smart on the client and correctly updated the voice state upon receiving a `user_muted` event but it looks like we are actually relying on an explicit voice off event on mute.

I guess we could keep doing the same but that would likely require a small addition on mobile so maybe we do it on this side although ultimately it's a bit of a wasted event since the state could be correctly inferred on client alone.

streamer45:
@cpoile This should be ready for final review, just in case you missed the notification.

cpoile (approved):
Nice fix!

PR #87, opened 2023-01-25
[MM-49980] Fix race condition on shutdown

streamer45:
#### Summary

PR fixes a potential race condition leading to a panic when shutting down the service.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-49980

### DIFF
--- a/service/ws/server_test.go
+++ b/service/ws/server_test.go
@@ -477,3 +477,33 @@ func TestSendMessages(t *testing.T) {
 
 	wg.Wait()
 }
+
+func TestRaceSendClose(t *testing.T) {
+	s, _, shutdown := setupServer(t)
+
+	var wg sync.WaitGroup
+	wg.Add(2)
+
+	go func() {
+		defer wg.Done()
+		for i := 0; i < 100; i++ {
+			_ = s.Send(Message{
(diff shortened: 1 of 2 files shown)


service/ws/server_test.go
```diff
@@ -477,3 +477,33 @@ func TestSendMessages(t *testing.T) {
...
+			})
+		}
+	}()
+
+	go func() {
+		defer wg.Done()
+		for i := 0; i < 100; i++ {
+			s.Close()
+		}
+		shutdown()
```
cpoile, line 503: Just for my own interest: I can't figure out what this is testing, exactly... :)

cpoile (approved):

service/ws/server_test.go
streamer45: @cpoile We are simply testing concurrent calls to `server.Close()` and `server.Send()` and iterating a while to make it more likely to trigger a race.

cpoile: I guess I'm wondering why the 100 `s.Close()`s? After the first one fires, the remaining `s.Send`s will silently error, right? (again, sorry, just trying to understand :) )

cpoile (approved):

PR #88, opened 2023-01-26
[MM-49998] Implement rate litiming on CPU heavy auth operations

streamer45:
#### Summary

PR adds some rate limiting to avoid the number of concurrent CPU heavy auth operations (mostly hashing) that could bring the instance to a halt if many clients connected concurrently.

I tried my best to come up with a sensible value (20 requests per second per CPU available) through some quick benchmarking but it may need more time to properly tune this.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-49998

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -20,6 +20,7 @@ require (
 	github.com/vmihailenco/msgpack/v5 v5.3.5
 	golang.org/x/crypto v0.2.0
 	golang.org/x/sys v0.2.0
+	golang.org/x/time v0.0.0-20191024005414-555d28b269f0
 )
 
 require (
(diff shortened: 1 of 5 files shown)


cpoile (approved):
Only suggestion might be: I've seen some code in the past that uses numCPU-1, to leave a core for the overhead. But since that's just a scaling factor here, it shouldn't matter.

streamer45:
> Only suggestion might be: I've seen some code in the past that uses numCPU-1, to leave a core for the overhead. But since that's just a scaling factor here, it shouldn't matter.

Indeed, it's just a multiplier at this point, we can't easily guarantee that one CPU would be left idle anyway since every request is a new goroutine.

cpoile (approved):

blue40524:
https://github.com/mattermost/rtcd/commit/9cd3f9f8826cdc653b75b1fb88613009ce34e2e2

PR #90, opened 2023-02-02
[MM-50288] Simulcast

streamer45:
#### Summary

PR adds simulcast support for screen sharing. The main functioning revolves around using a server-side bandwidth estimator to figure out if/when it makes sense to send a specific encoding to a client, basing the decision on both the estimated down rate for the receiving client and the actual up rate of the sending side.

Unfortunately after playing with the provided estimator from `pion/interceptor` for a while, I couldn't get stable results without making some small changes, hence [I had to fork](https://github.com/streamer45/interceptor) and import a modified `interceptor` package to allow for the target bitrate of the estimator to be tuned dynamically. (more context in https://github.com/pion/interceptor/issues/156)

I don't know how long we are going to support simulcast as it should likely be replaced by SVC once we get to implement that and once its behaviour is consistent across all major browsers but for now I think it will do. There are a few gotchas and the logic is far from perfect but I tried to balance as best as possible the sending of higher quality tracks vs avoiding data loss/congestion for clients unable to cope with the increased rate.

I understand there's a lot here as I also moved some code around so please ask me anything that may seem unclear :)

As a final note, while I tested this quite a bit already I am still planning to load-test it extensively before merging.

#### Related PR

https://github.com/mattermost/mattermost-plugin-calls/pull/316

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-50288


service/rtc/sfu.go
```diff
@@ -354,42 +408,93 @@ func (s *Server) InitSession(cfg SessionConfig, closeCb func() error) error {
...
 					)
 				}
 			})
 
+			limiter := rate.NewLimiter(0.25, 1)
+
 			for {
-				rtp, _, readErr := remoteTrack.ReadRTP()
+				buf := s.bufPool.Get().([]byte)
+				i, _, readErr := remoteTrack.Read(buf)
```
cpoile, line 470: iiuc, this is to get the number of bytes? Or is there another reason? If it's the former, then isn't something like `len(rtp.payload)` easier/faster?

service/rtc/simulcast.go
```diff
@@ -0,0 +1,210 @@
...
+func getRateForSimulcastLevel(level string) int {
+	var rate int
+	for idx, lvl := range simulcastLevels {
+		if lvl == level {
+			rate = simulcastRates[idx]
+			break
+		}
+	}
+	return rate
+}
```
cpoile, line 34: nit: Do we forsee adding more levels in the near future? If not, would it be simpler to do a check and return? (up to you)

service/rtc/stat/stat.go
```diff
@@ -0,0 +1,42 @@
+// Copyright (c) 2022-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+package stat
```
cpoile, line 4: I guess you looked at things like https://pkg.go.dev/github.com/montanaflynn/stats? (I thought we talked about that recently)

cpoile (approved):
Awesome, this is really good. Just little nits and comments, nothing too useful. :)

service/rtc/session.go
```diff
@@ -99,14 +112,47 @@ func (s *session) getScreenStreamID() string {
...
+		rid = SimulcastLevelDefault
+	}
+
+	return s.screenRateMonitors[rid]
+}
+
+func (s *session) getOutScreenTrack(rid string) *webrtc.TrackLocalStaticRTP {
+	s.mut.RLock()
+	defer s.mut.RUnlock()
+	return s.outScreenTracks[rid]
```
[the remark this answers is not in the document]
streamer45: The way it's currently used we don't because it's called in here:

https://github.com/mattermost/rtcd/blob/823fbd14146fce1a5ac7ec7980ac972064936209/service/rtc/simulcast.go#L166-L177

And this only applies in case of simulcast so the RID cannot be empty as `getSimulcastLevel()` will always return something.

service/rtc/sfu.go
```diff
@@ -83,10 +109,35 @@ func initInterceptors(m *webrtc.MediaEngine) (*interceptor.Registry, error) {
...
+	// TWCC
+	if err := webrtc.ConfigureTWCCSender(m, &i); err != nil {
+		return nil, nil, err
+	}
+
+	// Congestion Controller
+	bwEstimatorCh := make(chan cc.BandwidthEstimator, 1)
+	congestionController, err := cc.NewInterceptor(func() (cc.BandwidthEstimator, error) {
+		return gcc.NewSendSideBWE(
+			gcc.SendSideBWEInitialBitrate(int(float32(getRateForSimulcastLevel(SimulcastLevelLow))*0.50)),
```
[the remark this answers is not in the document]
streamer45: Originally I was starting with the medium quality track (we had three) mainly to avoid wasting too much time before upgrading to best quality (ideal case). But starting with high I think is a bit dangerous as it could cause bad congestion since at the start we don't know anything about the peer's network and can't asses its bandwidth. That said this is not set in stone, we can always review it later but I'd rather being on the safe side for now.

service/rtc/sfu.go
```diff
@@ -117,25 +168,19 @@ func (s *Server) InitSession(cfg SessionConfig, closeCb func() error) error {
 
 	peerConnConfig := webrtc.Configuration{
 		ICEServers:   iceServers,
-		SDPSemantics: webrtc.SDPSemanticsUnifiedPlanWithFallback,
+		SDPSemantics: webrtc.SDPSemanticsUnifiedPlan,
```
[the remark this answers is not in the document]
streamer45: Yeah, totally unrelated but good riddance ;)

service/rtc/sfu.go
streamer45: Yes, we need the number of received bytes to feed it to the monitor for calculating the source rate.

> If it's the former, then isn't something like len(rtp.payload) easier/faster?

That would only return the size of the payload, excluding any header and padding. To be accurate we want to pass everything we get since it's going to be relayed to the peers eventually.

service/rtc/simulcast.go
streamer45: Yeah. Originally we had three, which is why this logic is still more complex that how it needs to be. I can update it and if we ever want support more we implement it back.

service/rtc/stat/stat.go
streamer45: Actually I looked at https://pkg.go.dev/golang.org/x/perf/internal/stats but by then I already implemented everything. If we need something more complex I am happy to add a dependency but for now I am thinking it's fine.

cpoile (approved):
nicer 👍

PR #92, opened 2023-02-15
[MM-50511] Fix connectivity when no UDP address is configured

streamer45:
#### Summary

What can I say, it was a rabbit hole from e2e failures up to this. I was always concerned when updating dependencies and it turned out I had my good reasons. I honestly haven't had time to go into specifics of why connectivity broken but it's most likely due to https://github.com/pion/ice/pull/483.

As suggested by the author of the upstream changes, to listen on multiple interfaces we should now use [`MultiUDPMuxDefault`](https://github.com/pion/ice/blob/v2.3.0/udp_mux_multi.go#L18) in place of [`UDPMuxDefault`](https://github.com/pion/ice/blob/v2.3.0/udp_mux.go#L26). Of course there's the slight complication of doing socket sharding (opening multiple connections per same ip/port pair) so I had to rework the code a bit.

Since this was somewhat painful I decided to also update all other WebRTC dependencies to latest version since I'd rather test them once at this point as this will certainly need some soaking time.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-50511


service/rtc/server.go
```diff
@@ -94,77 +87,45 @@ func (s *Server) Start() error {
...
 		}
 
-		conns = append(conns, udpConn)
-	}
-	var err error
-	s.udpConn, err = newMultiConn(conns)
-	if err != nil {
-		return fmt.Errorf("failed to create multiconn: %w", err)
+		muxes = append(muxes, ice.NewUDPMuxDefault(ice.UDPMuxParams{
+			Logger:  newPionLeveledLogger(s.log),
```
streamer45, line 119: Turns out there was a warning printed but we didn't set any logger:

```
ice WARNING: 2023/02/14 18:19:24 UDPMuxDefault should not listening on unspecified address, use NewMultiUDPMuxFromPort instead
```

It's unlikely we would have caught this in the e2e tests pipeline but still something. Now passing a wrapper to our logger in order to potentially catch these things sooner in the future.

service/rtc/logger.go
```diff
@@ -0,0 +1,61 @@
...
+	log.log.Trace(fmt.Sprintf(format, args...))
+}
+
+func (log *pionLogger) Debug(msg string) {
+	log.log.Trace(msg)
+}
+
+func (log *pionLogger) Debugf(format string, args ...interface{}) {
+	log.log.Trace(fmt.Sprintf(format, args...))
+}
```
streamer45, line 37: Purposely downgrading these to trace since they are pretty spammy and we tend to default to DEBUG level for most operations.

cpoile: This is instead of changing the debugs -> trace for those that we actually just want traces of? (rather than removing the debug level)

cpoile (approved):
Thanks for the time spent on this!

service/rtc/logger.go
streamer45: If someone needs trace level logs they can still get them by changing the level in config. Also they can always run either Mattermost (if running with plugin) or `rtcd` with `PION_LOG_TRACE=ice` to get them as well.

PR #94, opened 2023-03-02
Improve local/external IPs mapping when listening on multiple interfaces

streamer45:
#### Summary

After changes in https://github.com/mattermost/rtcd/pull/92 there was an issue preventing connectivity under certain scenarios as  local (to the rtcd host) candidates could get dropped due to mapping to the same public IP (e.g. ice host override). The solution is to generate a set of `remoteIP/localIP` pairs and pass them down the pion stack.

Right now we are simply mapping the first available interface to the matching public IP and everything else to its own local IP if matching to the same public one. This fixes the connectivity issues I mentioned above as it lets candidates pass deduplication but it's not perfect especially since the first interface is usually the loopback one (e.g. 127.0.0.1). We could add a specific rule to avoid mapping this to a public address but I am a bit wary of changing the existing logic too much if nothing is broken.

Ideally we should surface this kind of configuration to let users deal with more advanced networking scenarios where multiple interfaces are actually used to deliver media to different sets of clients. Some potential improvement for the future as I don't want to complicate things further if there's no real need for it.

### DIFF
--- a/service/rtc/utils.go
+++ b/service/rtc/utils.go
@@ -4,29 +4,53 @@
 package rtc
 
 import (
-	"context"
 	"fmt"
-	"net"
 	"time"
 
 	"github.com/mattermost/rtcd/service/random"
 )
 
-func resolveHost(host string, timeout time.Duration) (string, error) {
-	var ip string
-	r := net.Resolver{}
-	ctx, cancel := context.WithTimeout(context.Background(), timeout)
-	defer cancel()
-	addrs, err := r.LookupIP(ctx, "ip4", host)
-	if err != nil {
-		return ip, fmt.Errorf("failed to resolve host %q: %w", host, err)
-	}
-	if len(addrs) > 0 {
-		ip = addrs[0].String()
-	}
-	return ip, err
-}
-
 func genTrackID(trackType, baseID string) string {
 	return trackType + "_" + baseID + "_" + random.NewID()[0:8]
+}
+
+func generateAddrsPairs(localIPs []string, publicAddrsMap map[string]string, hostOverride string) ([]string, error) {
+	var err error
+	var pairs []string
+	var hostOverrideIP string
+
+	if hostOverride != "" {
+		hostOverrideIP, err = resolveHost(hostOverride, time.Second)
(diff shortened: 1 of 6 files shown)


service/rtc/utils.go
```diff
@@ -4,29 +4,53 @@
...
+		publicAddr := publicAddrsMap[localAddr]
+
+		// If an override was explicitly provided we enforce that.
+		if hostOverrideIP != "" {
+			publicAddr = hostOverrideIP
+		}
+
+		if publicAddr != "" && !usedPublicAddrs[publicAddr] {
+			// if a public IP has not been used yet we map it to
+			// the first matching local ip.
```
cpoile, line 40: These comments were helpful, thank you

cpoile (approved):
Nice!

cpoile:
Oh, one maybe dumb question about:

> This fixes the connectivity issues I mentioned above as it lets candidates pass deduplication but it's not perfect especially since the first interface is usually the loopback one (e.g. 127.0.0.1). We could add a specific rule to avoid mapping this to a public address but I am a bit wary of changing the existing logic too much if nothing is broken.

Would this break anything though? It seems simple to add that rule... But maybe I don't see where could go wrong.

streamer45:
> Oh, one maybe dumb question about:
> 
> > This fixes the connectivity issues I mentioned above as it lets candidates pass deduplication but it's not perfect especially since the first interface is usually the loopback one (e.g. 127.0.0.1). We could add a specific rule to avoid mapping this to a public address but I am a bit wary of changing the existing logic too much if nothing is broken.
> 
> Would this break anything though? It seems simple to add that rule... But maybe I don't see where could go wrong.

That's a good question. I've seen very complicated setups with customers using calls so I wouldn't be surprised if there was a scenario where data that arrives through an external IP and somehow gets to the loopback interface via some special rules. So my approach here would be to avoid adding more "special" logic that could somewhat break things further. As things stand everything should work, regardless of the local IP mapped to the external one.

https://github.com/pion/ice/issues/541 has some more context which you may find helpful.

PR #95, opened 2023-03-10
Fix CPU leak

streamer45:
#### Summary

The `default` case would cause us to loop continuously causing a huge CPU resource leak.


cpoile (approved):
😬

PR #96, opened 2023-03-25
[MM-50288] Simulcast improvements

streamer45:
#### Summary

Adding some small improvements to the simulcast logic as I have been testing it more today. Most notable changes are:

- Added an additional check on average loss to determine whether to go ahead with the level change check. Essentially if there's no loss we don't want to downgrade, regardless of the estimated rate. Ideally this shouldn't be needed as the rate from the estimator should already account for that but especially with Firefox the rates are all over the place at times, resulting in a lot of downgrades under optimal network conditions.
- Improved source rate calculation. Also now using different sample sizes based on the simulcast level as samples for higher rate tracks will be much more frequent so we don't need as many for an accurate result.

Plan is to keep testing this for a while under different network conditions after which I'll conduct a load-test. I may send a few more changes if necessary but hopefully this should be the bulk of it.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-50288

### DIFF
--- a/service/rtc/rate.go
+++ b/service/rtc/rate.go
@@ -8,8 +8,6 @@ import (
 	"math"
 	"sync"
 	"time"
-
-	"github.com/mattermost/rtcd/service/rtc/stat"
 )
 
 type RateMonitor struct {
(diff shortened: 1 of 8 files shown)


service/rtc/rate.go
```diff
@@ -43,11 +41,14 @@ func (m *RateMonitor) PushSample(size int) {
 	m.mut.Lock()
 	defer m.mut.Unlock()
 
-	if !m.filled && m.getSamplesDuration() < m.samplingSize {
+	// Filling up to double the sampling size to make sure we have enough samples
+	// to calculate the desired duration since at the beginning it's likely we get
+	// a burst of packets.
+	if !m.filled && m.getSamplesDuration() < m.samplingSize*2 {
```
cpoile, line 47: I'm wondering if we should maybe give it some time to stabilize, and then start measuring?

cpoile (approved):
Thanks!

cpoile (approved):

streamer45:
Merging this. Will run another round of performance tests and if necessary will create a new PR to make additional tweaks.

PR #97, opened 2023-04-03
[MM-51861] Fix race condition with adding tracks on join

streamer45:
#### Summary

A race condition on join could cause tracks to be added more than once. This was spotted while I was reviewing some of the Prometheus metrics hence this PR includes those changes as well. In particular:

- Removed `rtp_packets_total` and `rtp_bytes_total`.
  - They caused a [ridiculous number of allocations](https://gist.github.com/streamer45/c6c098e3c0fb5057494054b9dd1642f3) as increment calls were called in a very tight loop.
  - They were not that useful to begin with since ultimately it's more important to track actual network rates.
  - With simulcast enabled, they were not going to be accurate any longer since tracks can be selectively added or removed based on clients' network capacity.
- Added `rtp_tracks_total` to keep count of input/output tracks and their type.
  - This is of particular importance since eventually it's the number one metric to predict performance. Ideally this should be used as the reference counter to load balance calls (tbd, related to https://github.com/mattermost/rtcd/pull/35).

Changes were load-tested.

image

Before merging I'll need to update our reference Grafana dashboard and documentation accordingly.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-51861

### DIFF
--- a/service/perf/metrics.go
+++ b/service/perf/metrics.go
@@ -19,11 +19,10 @@ const (
 type Metrics struct {
 	registry *prometheus.Registry
 
-	RTPPacketCounters      *prometheus.CounterVec
-	RTPPacketBytesCounters *prometheus.CounterVec
-	RTCSessions            *prometheus.GaugeVec
-	RTCConnStateCounters   *prometheus.CounterVec
-	RTCErrors              *prometheus.CounterVec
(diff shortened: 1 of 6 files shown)


cpoile (approved):
Nice and clean!

PR #98, opened 2023-04-03
Performance optimizations

streamer45:
#### Summary

The introduction of simulcast support in https://github.com/mattermost/rtcd/pull/90 caused the peak performance at scale, measured as CPU usage, to drop by roughly 50%. This was almost exclusively due to the bandwidth estimation and congestion control logic that was added to support the new functionality.

An extensive series of load-tests was conducted to try and identify any new bottlenecks. This and the related PR (where most performance related changes live) include a number of optimizations aimed at reducing overall CPU load at scale.

Note: currently using a temporary branch to keep the diff clean as this is based on top of both `MM-50288` and `MM-51861`.

#### Load-test info

 The load-test input parameters were the following:

- 25 concurrent calls
- 25 participants/call
- 2 unmuted participants/call
- 25 screen sharing senders (1 per call).

This generated 625 concurrent RTC sessions, 1200 outgoing voice tracks and 600 screen tracks with peak transmission bandwidth of 1.61Gb/s.

#### Results

The proposed changes so far could get back ~20% of the performance penalty of using the new interceptors.
Further work may be required but it may become harder to improve this by much as there's simply a lot more happening than before, and in tight loops (e.g. during RTP packets reception and handling).

calls_perf_optimizations_april_2023

Not applied here, but some potential improvement was seen by playing with the [`GOGC`](https://tip.golang.org/doc/gc-guide) variable which could be used to lower the garbage collector pressure at the expense of higher memory usage. Given the main bottleneck is CPU, it could make sense to consider this as an option for the future. Leaving here for reference some of results of such experiments:

calls_perf_optimizations_gogc_april_2023

#### Next steps

- Thoroughly test the proposed changes to make sure nothing was altered logic wise, especially with simulcasted tracks.
- Update the performance section in the official documentation (benchmarks) to reflect these results as the vertical scaling capabilities may have noticeably changed (https://mattermost.atlassian.net/browse/MM-51908)

#### Related PR

https://github.com/streamer45/interceptor/pull/2

#### Artifacts

Attaching the following artifacts:

- Baseline CPU and memory (heap) profiles.
- Pre-optimization CPU and memory (heap) profiles.
- Post-optimization CPU and memory (heap) profiles.

[calls_perf_april_2023.zip](https://github.com/mattermost/rtcd/files/11142973/calls_perf_april_2023.zip)

### DIFF
--- a/service/rtc/sfu.go
+++ b/service/rtc/sfu.go
@@ -98,7 +98,7 @@ func initInterceptors(m *webrtc.MediaEngine) (*interceptor.Registry, <-chan cc.B
 	}
 
 	// NACK
-	responder, err := nack.NewResponderInterceptor(nack.ResponderSize(nackResponderBufferSize))
+	responder, err := nack.NewResponderInterceptor(nack.ResponderSize(nackResponderBufferSize), nack.DisableCopy())
 	if err != nil {
 		return nil, nil, err
 	}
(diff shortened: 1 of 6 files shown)


service/rtc/sfu.go
```diff
@@ -492,15 +486,7 @@ func (s *Server) InitSession(cfg SessionConfig, closeCb func() error) error {
...
-				rtp := &rtp.Packet{}
-				if err := rtp.Unmarshal(buf[:i]); err != nil {
-					s.log.Error("failed to unmarshal RTP packet",
-						mlog.Err(err), mlog.String("sessionID", us.cfg.SessionID))
-					s.metrics.IncRTCErrors(us.cfg.GroupID, "rtp")
-					return
-				}
-
-				rm.PushSample(i)
+				rm.PushSample(packet.MarshalSize())
```
cpoile, line 488: That's smart

PR #99, opened 2023-04-17
Simulcast improvements

streamer45:
#### Summary

Some more improvements to the simulcast level switching logic after another batch of tests. 

PR also included some less related changes:

- Added delta profiles to support pulling data from [Pyroscope](https://pyroscope.io/).
- Added some integration tests to cover general connectivity to calls.


service/rtc/sfu.go
```diff
@@ -118,17 +118,15 @@ func initInterceptors(m *webrtc.MediaEngine) (*interceptor.Registry, <-chan cc.B
 	}
 
 	// Congestion Control
-	initialRate := int(float32(getRateForSimulcastLevel(SimulcastLevelLow)) * 0.50)
-	pacer, err := gcc.NewLeakyBucketPacer(initialRate, gcc.PacerDisableCopy())
-	if err != nil {
-		return nil, nil, fmt.Errorf("failed to create pacer: %w", err)
-	}
```
streamer45, line 125: This is a significant change I want to highlight. While I am still committed to improve the overall performance of the pacer I don't want to add too many changes at once so I am leaving this out for now. 

What this means in practical terms is that we still have bandwidth estimation in place which we require for simulcast but the congestion controller is not actuating the rate change so we are less protected to actual congestion. The NoOp pacer as name suggest sends out the packets directly without any buffering or artificial slow down.

cpoile (approved):
Nice, lots of work here, thank you!

streamer45:
Need to re-ask review, policies...

cpoile (approved):
There's no need for that policy on this repo, I thought they reverted that change?

streamer45:
> There's no need for that policy on this repo, I thought they reverted that change?

I think they did at some point but it's clearly back again.

PR #101, opened 2023-05-03
[MM-52650] Bump tracks channel size

streamer45:
#### Summary

The simplification introduced in https://github.com/mattermost/rtcd/pull/97 meant that on join we are queuing all the tracks before actually adding them. This means that if there are more tracks (e.g. participants that unmuted at least once throughout the call) than the capacity of the channel we'd fail to add some. 

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-52650

### DIFF
--- a/service/rtc/session.go
+++ b/service/rtc/session.go
@@ -22,7 +22,7 @@ import (
 
 const (
 	signalChSize = 20
-	tracksChSize = 10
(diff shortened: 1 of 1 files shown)


cpoile (approved):
Makes sense 👍

PR #113, opened 2023-07-06
[MM-53511] Properly remove screen sharing tracks if sender suddenly drops

streamer45:
#### Summary

In https://github.com/mattermost/rtcd/pull/107 I had a good thought of returning an error if a `screenTrackSender` was getting overwritten. The only problem was that it didn't prevent a panic from happening next time we tried to remove a `nil` track.

PR makes sure of two things:

1. (panic fix) If removing a track fails, we check whether that's because it was a `nil` track and we only extract its ID if that's not the case.
2. (functionality fix) When a session suddenly drops without sending the expected events (e.g. `ScreenOffMessage`) we are now properly removing the track from the receivers who are left and clearing `screenTrackSender` so that next time someone attempts to share it won't be failing.
 
#### Ticket Link

https://mattermost.atlassian.net/browse/MM-53511

### DIFF
--- a/service/rtc/call.go
+++ b/service/rtc/call.go
@@ -4,6 +4,7 @@
 package rtc
 
 import (
+	"fmt"
 	"sync"
 
 	"github.com/pion/webrtc/v3"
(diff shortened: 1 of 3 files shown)


cpoile (approved):
Awesome, thank you!

PR #115, opened 2023-07-12
[MM-53457] Standalone client

streamer45:
#### Summary

PR implements a standalone Golang client for Calls. This is mostly based of the [official one](https://github.com/mattermost/mattermost-plugin-calls/blob/main/webapp/src/client.ts) included in the plugin. The implementation abstracts away all the logic necessary to successfully join a call given a `SiteURL`, `AuthToken` and `ChannelID`.

This includes:
  - WebSocket client implementation.
  - WebRTC client implementation.
  - Any Calls plugin specific API logic (e.g. join message, reconnect support, etc). 

There are however some key differences:

- The WebRTC API is the one provided by [`pion`](https://github.com/pion/webrtc) which comes with its own set of limitations (e.g. perfect negotiation is not yet supported).
- The client is currently receive-only, purely useful for receiving tracks since that's the main use case. I don't exclude extending this to be a full functional client but not a priority. If we ever do that it may be interesting to replace the one used for load-testing with this.

This also comes with some additional limitations and potential TODOs:

- We are not reading the plugin's config which means we are not setting any ICE server configs. As the client is mostly meant to connect internally they shouldn't be necessary but something that could be added to make it a little closer to a real implementation.
- We are not implementing the MM reliable websocket functionality. This is partly because it doesn't really work in HA clusters so not sure if it's worth adding for this use case.

To note, this PR also includes a very basic e2e pipeline as tests are connecting against a running version of the plugin compiled with the latest `rtcd` changes. This could also serve as a good base to implement more tests that check general functionality.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-53457


cpoile (dismissed):
Looks very clean, nicely done. Nothing jumps out as a need to fix, and I need to get more familiar with pion to know if you're doing anything out of the ordinary there. But I'll assume you're not. 👍

streamer45:
Added a couple of extra commits to facilitate the transcription use case, mainly being able to react to user disconnecting and call ending.

client/config.go
```diff
@@ -19,6 +19,9 @@ type Config struct {
 	AuthToken string
 	// ChannelID is the id of the channel to start or join a call in.
 	ChannelID string
+	// ContextID is an id used to identify bot initiated sessions (e.g.
+	// recording/transcription)
+	ContextID string
```
cpoile, line 24: Just wondering if it can be named something more descriptive, to indicate what the ContextID means? I imagine I will forget what it means.

client/types.go
```diff
@@ -7,10 +7,16 @@ const pluginID = "com.mattermost.calls"
 
 type CallJoinMessage struct {
 	ChannelID string `json:"channelID"`
+	ContextID string `json:"contextID"`
```
cpoile, line 10: Likewise

cpoile (dismissed):
Nice, non-blocking comments

client/config.go
streamer45: For context (no pun intended :p), this is the same we are introducing in https://github.com/mattermost/mattermost-plugin-calls/pull/462 . The reason I kept it this abstract is that I could see more use cases in the future that don't necessarily map to what we are doing now. For the current purposes this could be easily called `JobID`, although it's a bit of an implementation details leak in my mind.

cpoile: If we do need to abstract it in the future, would it be easy to then abstract the field name?

streamer45: We'd need to either change the existing name (breaking change) or add a new field. Can be done, but could be costly.

cpoile: Personally I would prefer a more concrete name (since even now I've forgotten exactly what it holds), but your call. 👍

cpoile (dismissed):
Thanks!

jupenur (approved):

cpoile (approved):

PR #116, opened 2023-07-25
Improve logging

streamer45:
#### Summary

We were missing potentially useful warnings from our logs that would ease debugging issues. Given Pion has an easy way to pass a proper logger we wrap our existing one so that everything can get printed out nicely.

#### Sample output:

```
info  [2023-07-25 16:09:31.866 -06:00] peer connection state changed: connected      caller="rtc/logger.go:70" origin="webrtc/v3.(*PeerConnection).onConnectionStateChange webrtc/peerconnection.go:495"
warn  [2023-07-25 16:09:31.867 -06:00] test warning                                  caller="rtc/logger.go:78" origin="webrtc/v3.(*PeerConnection).undeclaredMediaProcessor webrtc/peerconnection.go:1627"
error [2023-07-25 16:09:31.867 -06:00] test error                                    caller="rtc/logger.go:86" origin="webrtc/v3.(*PeerConnection).undeclaredMediaProcessor webrtc/peerconnection.go:1628"
```

PR #117, opened 2023-07-25
[MM-53807] Explicitly stop receiver on exit

streamer45:
#### Summary

I wasn't able to reproduce this so not sure this addition will have any effect but until we get more logs (https://github.com/mattermost/rtcd/pull/116) it's hard to say what's going on. 

There's no clear indication that we are leaking anything on our side assuming that closing the peer connection would be enough to release all its resources.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-53807


cpoile (approved):
👍

PR #119, opened 2023-09-07
Fix broken dependency

streamer45:
#### Summary

Updating a dependency that was breaking the build when using the new Golang v1.21


cpoile (approved):
Why are we using go 1.21? Don't we usually stay a release behind?

PR #131, opened 2024-03-11
Fix linting on newer versions

streamer45:
#### Summary

Bumping the `github.com/grafana/pyroscope-go/godeltaprof` dependency as it's currently breaking `golangci-lint` on Go 1.22+

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -292,8 +292,7 @@ go-run: ## to run locally for development
 prepare-plugin: ## prepare Calls plugin for testing
 ifeq (${CI}, true)
 	@$(INFO) preparing Calls plugin...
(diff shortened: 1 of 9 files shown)


cpoile (approved):

streamer45:
@cpoile It was a bit painful but finally figured tests were failing because we were running an out-of-date MM server image which failed the recent v9.5 compatibility check so the plugin wouldn't start. It should be all fixed now but please take another look because the PR has gradually become something quite different.

cpoile (approved):

cpoile:
Nice, thanks for finding out the problem with that

PR #135, opened 2024-04-25
Avoid replacing track on mute

streamer45:
#### Summary

I was trying to follow what a real (browser) client would do but it looks like it's a bit bugged. As long as we pause writing on the track we should be fine.

### DIFF
--- a/client/api.go
+++ b/client/api.go
@@ -65,12 +65,6 @@ func (c *Client) Mute() error {
 	c.mut.Lock()
 	defer c.mut.Unlock()
 
-	if c.voiceSender != nil {
-		if err := c.voiceSender.ReplaceTrack(nil); err != nil {
-			return fmt.Errorf("failed to replace track: %w", err)
-		}
-	}
-
 	return c.sendWS(wsEventMute, nil, false)
 }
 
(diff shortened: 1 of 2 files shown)


cpoile (approved):
But does that really simulate what the client will be receiving when participants mute/unmute? I'm thinking about the client-side load of the pc replacing tracks.
I mean, we have no choice right now, I just hope we don't have to redo load tests once this is fixed upstream.

streamer45:
> But does that really simulate what the client will be receiving when participants mute/unmute? I'm thinking about the client-side load of the pc replacing tracks. I mean, we have no choice right now, I just hope we don't have to redo load tests once this is fixed upstream.

@cpoile Replacing tracks with `nil/null` is supposed to be a local operation with no server involvement (no negotiation). That's the whole point of doing it. So there's no impact on other clients whether we do it or not. 

Even then it's implementation-specific, for example, Firefox doesn't stop the sender while Chrome does.

See https://developer.mozilla.org/en-US/docs/Web/API/RTCRtpSender/replaceTrack for more details.

cpoile:
Awesome, thanks!

PR #137, opened 2024-05-07
[MM-58136] Support advanced mapping for ice host port override

streamer45:
#### Summary

PR adds support for a more advanced syntax in the `rtc.ice_host_port_override` config setting which takes the following form:

`localIPA/externalPortA,localIPB/externalPortB`

This becomes useful in deployments that are exposing a single public IP address and hence must be using different port numbers to map against multiple rtcd services listening internally. 

The reason to pass a full mapping is to make it possible to deploy in Kubernetes where the config passed must be the same across all the pods in the deployment. Each rtcd instance will only use the port override that maps to its local IP.

/cc @coltoneshaw 

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-58136

### DIFF
--- a/service/rtc/config.go
+++ b/service/rtc/config.go
@@ -7,6 +7,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net"
+	"strconv"
 	"strings"
 )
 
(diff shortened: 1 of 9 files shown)


service/rtc/config.go
```diff
@@ -206,3 +205,90 @@ func (s *ICEServers) UnmarshalTOML(data interface{}) error {
...
+		*s = ICEHostPortOverride(data.(string))
+		return nil
+	case int, int32, int64:
+		*s = ICEHostPortOverride(fmt.Sprintf("%v", data))
+	default:
+		return fmt.Errorf("unknown type %T", t)
+	}
+
+	return nil
+}
```
streamer45, line 294: This is to keep backward compatibility since this used to be an int before.

cpoile (approved):
Clean & nice, thank you!

streamer45:
Tested this on a two rtcd nodes deployment with a shared [NLB](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/introduction.html) in front. Looks good.

PR #138, opened 2024-05-08
[MM-54335] Improve write concurrency for screen tracks

streamer45:
#### Summary

Initial ceiling tests identified a bottleneck when writing screen tracks in large calls (> 250 participants). This is because the write loop runs on a single goroutine since we use a single output track. This results in the CPU not being utilized to the fullest and packets getting significantly delayed or even dropped. 

To overcome the current limitation, this PR proposes to create N output tracks and concurrently write to them. 

These changes are showing to improve overall performance and CPU utilization on our target instance (`c7i.xlarge`) by a factor of ~2.5, getting us from around 300 participants to 800 per call with near-zero packet loss.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-54335

### DIFF
--- a/service/rtc/sfu.go
+++ b/service/rtc/sfu.go
@@ -10,7 +10,6 @@ import (
 	"math/rand"
 	"os"
 	"runtime"
-	"sync"
 	"time"
 
 	"golang.org/x/time/rate"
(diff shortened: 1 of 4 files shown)


service/rtc/sfu.go
```diff
@@ -484,7 +488,21 @@ func (s *Server) InitSession(cfg SessionConfig, closeCb func() error) error {
...
+						genTrackID(trackTypeScreen, us.cfg.SessionID), random.NewID(), webrtc.WithRTPStreamID(remoteTrack.RID()))
+					if err != nil {
+						return nil, fmt.Errorf("failed to create screen track")
+					}
+					outTracks[i] = outTrack
+				}
+				return outTracks, nil
+			}
+
+			outScreenTracks, err := createOutScreenTracks(runtime.NumCPU())
```
streamer45, line 505: I am still debating whether `runtime.NumCPU()` is the optimal value. Using double that has shown even better CPU utilization but it comes with tradeoffs as one huge call could potentially bring the whole service to its knees, including other calls. So it may be better to be conservative to support a multi-call scenario.

service/rtc/sfu.go
```diff
@@ -564,11 +608,17 @@ func (s *Server) InitSession(cfg SessionConfig, closeCb func() error) error {
...
+				for i, writerCh := range writerChs {
+					// We need to copy the packet header to keep it race free in case
+					// of simulcast as we are dealing with concurrent writers.
+					pkt := *packet
+					pkt.Header = packet.Header.Clone()
+
+					select {
+					case writerCh <- &pkt:
+					default:
+						s.log.Error("failed to write RTP packet to writer channel", mlog.String("trackID", outScreenTracks[i].ID()))
```
streamer45, line 620: One unrelated positive thing about this approach is that if things start to go wrong the channel will become full and we'll log some errors. Before these changes, the process would simply struggle to send the packets in time but no error would show and I had to check on the client side for packet loss.

service/rtc/sfu.go
```diff
@@ -484,7 +487,21 @@ func (s *Server) InitSession(cfg SessionConfig, closeCb func() error) error {
 
 			s.log.Debug("received screen sharing stream", mlog.String("streamID", streamID), mlog.String("sessionID", us.cfg.SessionID))
 
-			outScreenTrack, err := webrtc.NewTrackLocalStaticRTP(params.RTPCodecCapability, genTrackID(trackTypeScreen, us.cfg.SessionID), random.NewID(), webrtc.WithRTPStreamID(remoteTrack.RID()))
+			// To improve concurrency and support larger calls we create NumCPU output tracks and randomly distribute the receivers among them.
+			createOutScreenTracks := func(num int) ([]*webrtc.TrackLocalStaticRTP, error) {
+				outTracks := make([]*webrtc.TrackLocalStaticRTP, num)
```
cpoile, line 492: I'm surprised you didn't get a go linting error on this -- they usually prefer a `make(.., 0, num)` and then append.

cpoile (approved):
Interesting, nice improvement!

service/rtc/sfu.go
cpoile: No I think I've had the linter say specifically for this case, it calls it a "c style" use of indexing. Afaik, appending after `make(..., 0, cap)` won't have a performance hit as long as you don't go over cap.

streamer45: Then I am sorry but that would be an absurd linter to have on :p

PR #140, opened 2024-05-16
[MM-58136] Apply port override to public addresses found through STUN

streamer45:
#### Summary

Covering another case which we'll need to unblock https://mattermost.atlassian.net/browse/CLD-7443. The port override should also apply to public addresses found through STUN since they effectively become host overrides.


cpoile (approved):
Thanks!

PR #142, opened 2024-05-22
Fix race in `CloseSession`

streamer45:
#### Summary

A test spotted a failure unrelated to the changes. Fixing the race seen in https://github.com/mattermost/rtcd/actions/runs/9198822361/job/25302371765?pr=141

Using a channel instead of the `sync.WaitGroup` is simpler.

### DIFF
--- a/service/rtc/call.go
+++ b/service/rtc/call.go
@@ -42,6 +42,7 @@ func (c *call) addSession(cfg SessionConfig, rtcConn *webrtc.PeerConnection, clo
 		sdpAnswerInCh:      make(chan webrtc.SessionDescription, signalChSize),
 		closeCh:            make(chan struct{}),
 		closeCb:            closeCb,
+		doneCh:             make(chan struct{}),
 		tracksCh:           make(chan trackActionContext, tracksChSize),
 		outScreenTracks:    make(map[string][]*webrtc.TrackLocalStaticRTP),
 		remoteScreenTracks: make(map[string]*webrtc.TrackRemote),
--- a/service/rtc/server_test.go
+++ b/service/rtc/server_test.go
@@ -248,12 +248,17 @@ func TestInitSession(t *testing.T) {
 		}
 	}
 
+	var wg sync.WaitGroup
+	wg.Add(len(sessions))
 	for _, cfg := range sessions {
 		go func(cfg SessionConfig) {
+			defer wg.Done()
 			err := s.InitSession(cfg, nil)
 			require.NoError(t, err)
 		}(cfg)
 	}
+
+	wg.Wait()
 
 	for _, cfg := range sessions {
 		go func(id string) {
--- a/service/rtc/session.go
+++ b/service/rtc/session.go
@@ -51,7 +51,7 @@ type session struct {
 
 	closeCh chan struct{}
 	closeCb func() error
-	doneWg  sync.WaitGroup
+	doneCh  chan struct{}
 
 	vadMonitor *vad.Monitor
 
--- a/service/rtc/session_test.go
+++ b/service/rtc/session_test.go
@@ -50,6 +50,8 @@ func TestAddSession(t *testing.T) {
 		require.NoError(t, err)
 		require.NotNil(t, us)
 
+		close(us.doneCh)
+
 		err = server.CloseSession(cfg.SessionID)
 		require.NoError(t, err)
 	})
@@ -80,6 +82,8 @@ func TestAddSession(t *testing.T) {
 		require.NoError(t, err)
 		require.NotNil(t, us)
 
+		close(us.doneCh)
+
 		err = server.CloseSession(cfg.SessionID)
 		require.Error(t, err)
 		require.Equal(t, cbError, err)
@@ -87,6 +91,8 @@ func TestAddSession(t *testing.T) {
 		us, err = server.addSession(cfg, peerConn, closeCbSuccess)
 		require.NoError(t, err)
 		require.NotNil(t, us)
+
+		close(us.doneCh)
 
 		err = server.CloseSession(cfg.SessionID)
 		require.NoError(t, err)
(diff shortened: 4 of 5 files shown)


cpoile (approved):
I'm a little unsure why this is simpler though...?
And is it or is it not related to the test failure? If so, I'm not sure I get how or why.

streamer45:
> I'm a little unsure why this is simpler though...? And is it or is it not related to the test failure? If so, I'm not sure I get how or why.

@cpoile Well, simpler is relative I suppose. In this case, a channel fits best because it's thread-safe by design.
The problem we had before that caused the test failure was trying to close a session before `InitSession()` got a chance to complete. That would cause us to wait on the wait group while concurrent adding on it. With a channel that problem goes away.

cpoile:
Hmm... I thought waitGroup was also thread safe by design... Would love to go through this a bit maybe on 1:1 to dig deeper. :)

streamer45:
> Hmm... I thought waitGroup was also thread safe by design... Would love to go through this a bit maybe on 1:1 to dig deeper. :)

@cpoile I think docs in https://pkg.go.dev/sync#WaitGroup.Add sums it up well.

cpoile:
Ah, but that's different than threadsafe, it's more like a sequencing or out of order issue, isn't it? The only reason the `us.doneCh` works is that you create when creating the session, and the reason waitgroup didn't work because you add to it later.
Just trying to understand if I had waitgroup wrong in my head.

streamer45:
> Ah, but that's different than threadsafe, it's more like a sequencing or out of order issue, isn't it? The only reason the `us.doneCh` works is that you create when creating the session, and the reason waitgroup didn't work because you add to it later. Just trying to understand if I had waitgroup wrong in my head.

@cpoile Initializing the channel with the session is what makes it simpler (IMO) but operations on a wait group are not always thread-safe as the failing test demonstrated. You cannot call `wg.Wait()` and `wg.Add()` concurrently on a fresh wait group because. Adding must happen first. Otherwise, it's racy, regardless of any sequencing issues.


Take this example:

```go
package main

import (
	"sync"
)

func main() {
	var wg sync.WaitGroup

	go func() {
		wg.Add(1)
		wg.Done()
	}()

	go func() {
		wg.Wait()
	}()
}
```

`go run -race main.go`

```
==================
WARNING: DATA RACE
Write at 0x00c00009c008 by goroutine 7:
  runtime.racewrite()
      <autogenerated>:1 +0x1e
  main.main.func2()
      /home/streamer45/code/go/wg_race/main.go:16 +0x2e

Previous read at 0x00c00009c008 by goroutine 6:
  runtime.raceread()
      <autogenerated>:1 +0x1e
  main.main.func1()
      /home/streamer45/code/go/wg_race/main.go:11 +0x33

Goroutine 7 (running) created at:
  main.main()
      /home/streamer45/code/go/wg_race/main.go:15 +0x125

Goroutine 6 (running) created at:
  main.main()
      /home/streamer45/code/go/wg_race/main.go:10 +0xbc
==================
Found 1 data race(s)
exit status 66
```

cpoile:
I get that, and I see why that example is racy -- but I don't think that's because of thread safety, that's because of sequencing (aka, a race condition). Maybe my conception of thread safety is wrong, but in my head it means that calling the same method or field from separate threads will never result in an invalid state. E.g., the built-in map is not totally threadsafe. But a waitgroup is -- you can call whatever you want with your waitgroup across threads and nothing will ever crash. But you have to `add` before you `wait` obviously, or else the `wait` won't wait, you know?

streamer45:
> I get that, and I see why that example is racy -- but I don't think that's because of thread safety, that's because of sequencing (aka, a race condition). Maybe my conception of thread safety is wrong, but in my head it means that calling the same method or field from separate threads will never result in an invalid state. E.g., the built-in map is not totally threadsafe. But a waitgroup is -- you can call whatever you want with your waitgroup across threads and nothing will ever crash. But you have to `add` before you `wait` obviously, or else the `wait` won't wait, you know?

@cpoile It can't be thread-safe if it can cause a race condition when running concurrently, regardless of whether it causes a crash or not. In this case it's the use that determines whether it's safe or not. So we could say that a wait group is thread-safe if you follow its rules. But we weren't :)

It's the same reason you cannot assign the same variable from two goroutines if they run concurrently, but it's totally safe to do it if you synchronize them properly.

cpoile:
This is a good conversation, but I don't want to take up too much time on it. Sorry for rabbit holing on it. I do think that you're using the word "threadsafe" differently from how I've always thought it to mean. Threadsafety (how I understand it) isn't about how you use the datastructure, rather it's just about whether or not you can access it and/or modify it from multiple threads concurrently. E.g., the racy example you gave is racy, but its raciness has nothing to do with threadsafety. A waitgroup is totally threadsafe, but if you use it incorrectly (like in the example) it will be racy. 
Again, maybe I'm using the term threadsafe wrong. But... I don't think so... 🤔

streamer45:
> This is a good conversation, but I don't want to take up too much time on it. Sorry for rabbit holing on it. I do think that you're using the word "threadsafe" differently from how I've always thought it to mean. Threadsafety (how I understand it) isn't about how you use the datastructure, rather it's just about whether or not you can access it and/or modify it from multiple threads concurrently. E.g., the racy example you gave is racy, but its raciness has nothing to do with threadsafety. A waitgroup is totally threadsafe, but if you use it incorrectly (like in the example) it will be racy. Again, maybe I'm using the term threadsafe wrong. But... I don't think so... 🤔


@cpoile This is a great discussion. I am happy to continue at our own pace.

> Threadsafety (how I understand it) isn't about how you use the datastructure, rather it's just about whether or not you can access it and/or modify it from multiple threads concurrently.

You can always access memory concurrently, but is it safe to do it? It's safe only as long as you don't cause a race. 

More in general, how would you classify a program that makes improper use of waitgroup? Would the program be considered thread-safe?

Maybe tying thread safety to a particular data structure isn't the best way to explain the problem here. The final result is what I really care about.

PR #150, opened 2024-06-24
[MM-52657] AV1 support

streamer45:
#### Summary

PR adds support for receiving and sending [AV1](https://en.wikipedia.org/wiki/AV1)-encoded screen-sharing tracks.

A new `av1Support` session property is used to decide whether we should send the AV1 track or the fallback VP8. This ensures that functionality on either old or supporting clients (e.g., mobile) can still function as expected.

Changes on this side should be backward compatible with previous plugin versions.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-52657


service/rtc/session.go
```diff
@@ -145,11 +145,11 @@ func (s *session) getSourceRate(rid string) int {
 	return rate
 }
 
-func (s *session) getOutScreenTrack(rid string) *webrtc.TrackLocalStaticRTP {
+func (s *session) getOutScreenTrack(mimeType, rid string) *webrtc.TrackLocalStaticRTP {
 	s.mut.RLock()
 	defer s.mut.RUnlock()
 
-	return pickRandom(s.outScreenTracks[rid])
+	return pickRandom(s.outScreenTracks[getTrackIndex(mimeType, rid)])
```
cpoile, line 152: Just curious, why is it pickRandom?

cpoile (approved):

service/rtc/session.go
streamer45: @cpoile Because of https://github.com/mattermost/rtcd/pull/138 if you recall

PR #151, opened 2024-07-18
[MM-59548] Fix potentially stuck goroutines

streamer45:
#### Summary

Looking at goroutines profiles highlighted a couple of (unrelated) issues:

1. Some read calls on incoming voice tracks can suddenly block due to an apparent regression upstream (`pion/transport`). 	  	
	- This is still under investigation. Downgrading to the previous version is enough to prevent the issue while a fix is found.
2. The `CloseSession` could get stuck indefinitely in case of a timeout during the initial signaling phase due to a circular check on the session's done channel.
	- Reworked the function to preemptively close the channel in these fail-early cases.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-59548

### DIFF
--- a/client/api_test.go
+++ b/client/api_test.go
@@ -717,3 +717,162 @@ func TestAPIConcurrency(t *testing.T) {
 		}
 	})
 }
+
+func TestAPIScreenShareAndVoice(t *testing.T) {
+	th := setupTestHelper(t, "calls0")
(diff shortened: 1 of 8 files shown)


cpoile (approved):
Huh, tricky. Nice fix. :tada:

PR #153, opened 2024-08-15
Fix CI

streamer45:
#### Summary

There were a couple of issues:

- The deprecated `docker-compose` command was used. This also meant the MM container name used for e2e tests had to be updated.
- Since we didn't fetch git tags, the version in the manifest wasn't generated properly causing a parsing failure.
- Testing against the plugin would fail as we merged the v1 breaking changes to limit calls to DMs only.

### DIFF
--- a/build/test/prepare-plugin.sh
+++ b/build/test/prepare-plugin.sh
@@ -1,5 +1,5 @@
 #/bin/bash
-set -x
+set -xe
 
 GIT_DEFAULT_BRANCH="main"
 GIT_REPO="https://github.com/mattermost/mattermost-plugin-calls"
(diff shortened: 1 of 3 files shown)


build/test/prepare-plugin.sh
```diff
@@ -14,27 +14,29 @@ else
...
 PLUGIN_BUILD_PATH=$(realpath dist/*.tar.gz)
 PLUGIN_FILE_NAME=$(basename ${PLUGIN_BUILD_PATH})
 
-docker cp ../rtcd/build/test/config_patch.json mmserver_server_1:/mattermost && \
-docker exec mmserver_server_1 bin/mmctl --local config patch config_patch.json && \
-docker cp ${PLUGIN_BUILD_PATH} mmserver_server_1:/mattermost && \
-docker exec mmserver_server_1 bin/mmctl --local plugin delete ${PLUGIN_ID} && \
-docker exec mmserver_server_1 bin/mmctl --local plugin add ${PLUGIN_FILE_NAME} && \
-docker exec mmserver_server_1 bin/mmctl --local plugin enable ${PLUGIN_ID} && \
+docker ps -a && \
```
cpoile, line 30: Is this for testing, or needed?

cpoile (approved):
Thanks!

build/test/prepare-plugin.sh
streamer45: Let's say I wish I had it before going through this. It doesn't hurt as it adds more visibility into the pipeline.

PR #154, opened 2024-08-15
[MM-60189] Rate limit forwarding PLI requests

streamer45:
#### Summary

Sending keyframes can be expensive for the presenter, as it requires more bandwidth. It gets worse when N receivers concurrently ask for a keyframe. This is common when a presenter begins screen sharing in a large call.

PR introduces some rate limiting to avoid forwarding more than 1 PLI (picture loss indication) request per second.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-60189

### DIFF
--- a/client/rtc.go
+++ b/client/rtc.go
@@ -256,8 +256,9 @@ func (c *Client) initRTCSession() error {
 
 		// RTCP handler
 		go func(rid string) {
+			var err error
+			rtcpBuf := make([]byte, receiveMTU)
 			for {
-				rtcpBuf := make([]byte, receiveMTU)
 				if rid != "" {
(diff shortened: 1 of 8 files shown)


client/rtc.go
```diff
@@ -242,14 +243,22 @@ func (c *Client) initRTCSession() error {
...
+		}
+
 		c.mut.Lock()
 		c.receivers[sessionID] = append(c.receivers[sessionID], receiver)
 		c.mut.Unlock()
 
 		// RTCP handler
 		go func(rid string) {
+			var err error
+			rtcpBuf := make([]byte, receiveMTU)
```
cpoile, line 260: yikes, good catch.

cpoile (approved):
Nice solution!

PR #155, opened 2024-08-22
[MM-60256] Support for a FQDN as ice host candidate override

streamer45:
#### Summary

PR adds experimental support for forwarding [FQDNs](https://en.wikipedia.org/wiki/Fully_qualified_domain_name) in the address field of the ICE host candidate. This can be particularly useful in scenarios where the client-side resolution can determine connectivity, such as VPNs or load balancers that would naturally return multiple IPs.

image

Now, of course, these don't seem to work in Firefox, hence the do-not-merge label :persevere: 

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-60256

### DIFF
--- a/config/config.sample.toml
+++ b/config/config.sample.toml
@@ -77,6 +77,11 @@ ice_host_override = ""
 # config to multiple pods in Kubernetes deployments. In that case, each pod should match against one
 # local (node) IP and greatly simplify load balancing across multiple nodes.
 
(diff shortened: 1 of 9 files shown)


streamer45:
> Now, of course, these don't seem to work in Firefox, hence the do-not-merge label 😣

https://bugzilla.mozilla.org/show_bug.cgi?id=1713128 :unamused:

PR #158, opened 2024-10-02
DC messages

streamer45:
#### Summary

Building on top of the previous PR (https://github.com/mattermost/rtcd/pull/156), we introduce a simple (msgpack based) format for messages exchanged through the data channel. The idea is to support a variety of different messages in the future, including some [QoE metrics](https://mattermost.atlassian.net/browse/MM-60561) coming from clients.


cpoile (approved):
:+1:

PR #160, opened 2024-10-10
Review histograms buckets

streamer45:
#### Summary

PR reviews the buckets for our histograms since the [default ones](https://github.com/prometheus/client_golang/blob/05fcde9fe4eb93d3fb7b56ebe51acf80536f0583/prometheus/histogram.go#L265C28-L265C64) didn't make too much sense with the range of values we'd expect. For example, range for loss rate is between 0 and 1 so we don't need further buckets.

The same goes for latency values such as round-trip times that we cap at 1 second because past that it's almost pointless to track as the app would be unusable.

### DIFF
--- a/service/perf/metrics.go
+++ b/service/perf/metrics.go
@@ -17,6 +17,11 @@ const (
 	metricsSubSystemWS        = "ws"
 )
 
+var (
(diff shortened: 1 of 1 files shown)


cpoile (approved):
:+1:

PR #161, opened 2024-11-01
[MM-61426] Migrate to pion v4

streamer45:
#### Summary

PR migrates to the new `pion/webrtc` v4 and related dependencies. 

Of note, I conducted a quick load test to ensure performance was on par, and it didn't go as expected (20-30% worse). After some digging, I discovered the issue being on the client side, meaning the updated client is now sending significantly more (one order of magnitude) RTCP messages ([twcc](https://bloggeek.me/webrtcglossary/transport-cc/) especially), which are causing additional load on the server. Disabling that interceptor brings the performance back to the expected level. 

What this means is that the server performs equally (less than 5% variance) when load-tested by a non-updated client. I'll need to review this separately because I don't yet know how this new client behaviour compares to a real client (e.g. web browser) (https://mattermost.atlassian.net/browse/MM-61476) but the backend update can proceed.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-61426


client/rtc.go
```diff
@@ -191,15 +191,17 @@ func (c *Client) initRTCSession() error {
...
-	statsInterceptorFactory, err := stats.NewInterceptor()
-	if err != nil {
-		return fmt.Errorf("failed to create stats interceptor: %w", err)
-	}
 	var statsGetter stats.Getter
-	statsInterceptorFactory.OnNewPeerConnection(func(_ string, g stats.Getter) {
-		statsGetter = g
-	})
-	i.Add(statsInterceptorFactory)
+	if c.cfg.EnableRTCMonitor {
```
cpoile, line 195: 👍

cpoile (approved):
Seems simple enough, thanks.
Nice work on performance testing!

client/rtc.go
streamer45: Yes, as part of my investigation I wanted to rule out this wasn't a problem so it made sense to only register this interceptor when the monitor is enabled. Either way, it ended up not being a perf issue.

streamer45:
This will take longer as there's another regression with how we track metrics for outgoing tracks due to how the connection closing logic changed.

PR #163, opened 2024-12-18
[MM-62310] Fix handling of empty RTP payloads with padding

streamer45:
image
Yeah, it's a very small thing, but during debugging [I used ChatGPT](https://chatgpt.com/c/67632931-d520-800b-b3a1-b658ac921b48) to quickly decode the RTP packets, which was useful since it would have taken me a few more minutes to code something.

#### Summary

So, this little thing caused transcriber jobs to fail at https://github.com/mattermost/mattermost-plugin-calls/pull/928.

Since the packets have no payload we can safely ignore any padding since it will get stripped by the pion layer.

I spent 1h+ working on an integration test, but it's very tricky to trigger since we don't have access to the lower level, and pion (since it's bugged) will fail to unmarshal these packets on the rtcd side, before we can get them to the receiving side and check for error. 

Having our e2e tests pipeline green again would be a strong confirmation this is fixed.

Note: tests will fail on this side until we sync with the above PR on the plugin side.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-62310

### DIFF
--- a/service/rtc/sfu.go
+++ b/service/rtc/sfu.go
@@ -488,6 +488,15 @@ func (s *Server) InitSession(cfg SessionConfig, closeCb func() error) error {
 					return
 				}
 
+				// Mitigating against https://github.com/pion/webrtc/issues/2403
+				// The padding will be stripped by pion but the header bit will be forwarded as is (set).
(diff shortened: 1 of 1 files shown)


cpoile (approved):
Good comment, thanks!

PR #164, opened 2024-12-19
Enable SCTP optimization on client side

streamer45:
#### Summary

> EnableSCTPZeroChecksum controls the zero checksum feature in SCTP. This removes the need to checksum every incoming/outgoing packet and will reduce latency and CPU usage.


cpoile (approved):
👍

PR #166, opened 2025-01-08
[MM-62410] Implement API handler to fetch session config

streamer45:
#### Summary

PR implements a new `/calls/{callID}/sessions/{sessionID}` HTTP API endpoint to let clients access information about a specific session (its config). This will enable the plugin side to know whether a session exists and help deal with potential inconsistencies (e.g., stuck calls).

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-62410

### DIFF
--- a/service/client.go
+++ b/service/client.go
@@ -11,9 +11,11 @@ import (
 	"log"
 	"net"
 	"net/http"
+	"net/url"
 	"sync"
 	"time"
 
+	"github.com/mattermost/rtcd/service/rtc"
 	"github.com/mattermost/rtcd/service/ws"
 )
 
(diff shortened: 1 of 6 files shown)


service/client.go
```diff
@@ -38,6 +40,52 @@ type Client struct {
 	wg  sync.WaitGroup
 }
 
+type ClientConfig struct {
```
streamer45, line 43: This code was simply moved from `config.go` since it made more sense to keep it closer to the rtcd Client implementation.

streamer45:
Added a new handler to get all the sessions for a given call since it's a little more efficient for the current use case.

PR #173, opened 2025-03-24
[MM-63430] Calls 1-1 video support (MVP)

streamer45:
#### Summary

PR includes changes that enable clients to share video tracks simultaneously.

There's nothing fundamentally new other than the `MediaMap` logic (see https://github.com/mattermost/calls-common/pull/47 for context). Other than that, we are mostly adapting what we were already doing with voice tracks but extended to video.

We also include Go client changes (e.g. used by our load-test tool).

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-63430


service/rtc/sfu.go
```diff
@@ -550,9 +567,15 @@ func (s *Server) InitSession(cfg SessionConfig, closeCb func() error) error {
...
-			us.screenRateMonitors[trackIdx] = rm
+			if isScreen {
+				us.outScreenTracks[trackIdx] = outTracks
+				us.remoteScreenTracks[trackIdx] = remoteTrack
+				us.screenRateMonitors[trackIdx] = rm
+			} else {
+				us.outVideoTracks[trackIdx] = outTracks
+				us.remoteVideoTracks[trackIdx] = remoteTrack
+				us.videoRateMonitors[trackIdx] = rm
+			}
```
cpoile, line 578: Is there a way to remove the duplication?

service/rtc/sfu.go
```diff
@@ -592,13 +615,19 @@ func (s *Server) InitSession(cfg SessionConfig, closeCb func() error) error {
 				)
 
 				select {
-				case ss.tracksCh <- trackActionContext{action: trackActionAdd, track: pickRandom(outScreenTracks)}:
+				case ss.tracksCh <- trackActionContext{
+					action:        trackActionAdd,
+					localTrack:    pickRandom(outTracks),
```
cpoile, line 620: I don't think I fully understand why `pickRandom` is used here

service/rtc/sfu.go
cpoile, line 578: Is this duplication that could be fixed?

service/rtc/session.go
```diff
@@ -62,11 +62,19 @@ type session struct {
...
+	outScreenTracks        map[string][]*webrtc.TrackLocalStaticRTP
+	outScreenAudioTrack    *webrtc.TrackLocalStaticRTP
+	remoteScreenAudioTrack *webrtc.TrackRemote
+	remoteScreenTracks     map[string]*webrtc.TrackRemote
+	screenRateMonitors     map[string]*RateMonitor
+	// Video
+	videoStreamID     string
+	outVideoTracks    map[string][]*webrtc.TrackLocalStaticRTP
+	remoteVideoTracks map[string]*webrtc.TrackRemote
+	videoRateMonitors map[string]*RateMonitor
```
cpoile, line 77: I'm also wondering about duplication here (screen vs video tracks). Maybe it's unavoidable?

service/rtc/track.go
```diff
@@ -15,6 +15,8 @@ const (
 )
 
 type trackActionContext struct {
-	action trackAction
-	track  webrtc.TrackLocal
+	action        trackAction
+	localTrack    webrtc.TrackLocal
+	remoteTrack   *webrtc.TrackRemote
+	senderSession *session
```
cpoile, line 21: Could I get an explanation for the change here (which is throughout, really) -- since the localTrack and remoteTrack differentiation is new, I'm just wondering why we didn't need it before and why we need it now? (I think my context window for the sfu blew past my token limit a little while ago)

cpoile (commented):
Awesome, thank you! just questions for my own understanding really.

service/rtc/sfu.go
streamer45: This is nothing new. Just applying the same optimization we use for screen tracks since they are still comparable in terms of type and rate. See https://github.com/mattermost/rtcd/pull/138

service/rtc/track.go
streamer45: Good question. These are useful so we can avoid a bunch of complexity in `handleSenderRTCP`. For screen sharing, it was simple enough to track things on the `call` object because we had the guarantee of one screen track/session globally per call at any given time, but now we could be handling concurrent incoming video tracks so it's a bit simpler to do it this way since we only need this info in there.

https://github.com/mattermost/rtcd/blob/7bc1f7ac6460f11520335e60f075a0acd9aa3f02/service/rtc/session.go#L330-L351

service/rtc/sfu.go
cpoile: ahhh, I forgot about that. Gotcha, thanks.

service/rtc/sfu.go
cpoile: Cool, I just want to make sure we're not missing something actually beneficial. Definitely don't needlessly generalize.

service/rtc/track.go
cpoile: Ah, I see. Yep, good idea to keep it separate, I like it.

streamer45:
Plan is to prepare a dev release from this PR's branch so we can test in Community without having to merge these changes to master, which would definitely require a bunch of extra reviews (Security + QA).

streamer45:
Publishing a `v1.2.0-dev0` tag out of this branch.

bgardner8008:
I've rebased the video changes onto master to get recent changes. It seems we've been running the older video branch on both hub and community production servers for some time, as evidenced by RTCD emitting MediaMap messages, which are specific to video branch, and logged as unknown message by non-video clients. So there's a high confidence this code works for non-video calls.

PR #175, opened 2025-04-18
[MM-63776] Dynamic video encodings

streamer45:
#### Summary

Matching server-side PR to https://github.com/mattermost/calls-common/pull/51

Left some comments to aid the review.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-63776

### DIFF
--- a/service/rtc/call.go
+++ b/service/rtc/call.go
@@ -63,7 +63,23 @@ func (c *call) addSession(cfg SessionConfig, rtcConn *webrtc.PeerConnection, clo
 	}
 
 	c.sessions[cfg.SessionID] = s
+
+	c.sendCodecSupportMap(log)
+
 	return s, true
+}
+
+// sendCodecSupportMap sends the codec support map to all sessions.
+// NOTE: this is expected to always be called under lock (call.mut).
+func (c *call) sendCodecSupportMap(log mlog.LoggerIFace) {
+	supportMap := c.getCodecSupportMap()
+	for _, us := range c.sessions {
+		select {
+		case us.dcOutCh <- dcMessage{msgType: dc.MessageTypeCodecSupportMap, payload: supportMap}:
+		default:
+			log.Error("failed to send codec support map: channel is full", mlog.String("sessionID", us.cfg.SessionID))
+		}
+	}
 }
 
 func (c *call) getScreenSession() *session {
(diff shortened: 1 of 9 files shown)


service/rtc/call.go
```diff
@@ -63,9 +64,25 @@ func (c *call) addSession(cfg SessionConfig, rtcConn *webrtc.PeerConnection, clo
 	}
 
 	c.sessions[cfg.SessionID] = s
+
+	c.sendCodecSupportMap(log)
```
streamer45, line 68: This is where we send the updated map after a new session joins. It's important that session also gets it, so this call must happen after updating the sessions map (line above).

service/rtc/session.go
```diff
@@ -77,9 +77,9 @@ type session struct {
 	videoRateMonitors map[string]*RateMonitor
 
 	// Receiver
-	bwEstimator       cc.BandwidthEstimator
-	screenTrackSender *webrtc.RTPSender
-	rxTracks          map[string]webrtc.TrackLocal
+	bwEstimator        cc.BandwidthEstimator
+	screenTrackSenders map[string]*webrtc.RTPSender
```
streamer45, line 81: Now we have to support potentially concurrent screen tracks, one per encoding (VP8 and AV1 at the moment).

service/rtc/session.go
```diff
@@ -417,9 +417,9 @@ func (s *session) addTrack(sdpOutCh chan<- Message, localTrack webrtc.TrackLocal
 		}
 	}
 
-	if getTrackType(localTrack.ID()) == trackTypeScreen && s.screenTrackSender != nil {
+	if getTrackType(localTrack.ID()) == trackTypeScreen && s.screenTrackSenders[getTrackMimeType(localTrack)] != nil {
```
streamer45, line 420: Still, no more than one track per encoding is expected (e.g. can't have 2 AV1 screen tracks).

service/rtc/sfu.go
```diff
@@ -762,6 +767,8 @@ func (s *Server) CloseSession(sessionID string) error {
 			s.mut.Unlock()
 		}
 		group.mut.Unlock()
+	} else {
+		call.sendCodecSupportMap(s.log)
```
streamer45, line 771: No need to send the map if the call is ending.

service/rtc/sfu.go
```diff
@@ -780,6 +787,58 @@ func (s *Server) CloseSession(sessionID string) error {
 	return nil
 }
 
+// collectSessionTracks gathers all available tracks from a session that should be delivered to the target session.
+func collectSessionTracks(senderSession *session, targetSession *session) ([]*webrtc.TrackLocalStaticRTP, []*webrtc.TrackRemote) {
```
streamer45, line 791: This is to address https://github.com/mattermost/rtcd/pull/173#discussion_r2025176872, especially since now it gets even more complicated. Still a somewhat involved method but a little better I suppose.

service/rtc/simulcast.go
```diff
@@ -158,7 +158,8 @@ func (s *session) handleSenderBitrateChange(downRate int, lossRate int) (bool, i
 	}
 
 	s.mut.RLock()
-	sender := s.screenTrackSender
+	// For now we only support simulcast on VP8.
+	sender := s.screenTrackSenders[webrtc.MimeTypeVP8]
```
streamer45, line 162: This should be safe, considering we disable AV1 on the client when simulcast is on. Also, we may deprecate simulcast soon if we don't fix it for good.

PR #176, opened 2025-05-26
[MM-64239] Extend `ICEAddressUDP` and `ICEAddressTCP` to support a specified list of addresses

streamer45:
#### Summary

Listening on the loopback device (i.e., `127.0.0.1`) is not always ideal and can sometimes cause issues — we've received a few reports about this. Typically, setting `ICEAddressUDP` and `ICEAddressTCP` to the IP address of the appropriate local network interface (e.g., `eth0`) is sufficient.

However, complications arise when there are multiple valid local interfaces — for example, one for the local network and another for VPN traffic. In such cases, we’re currently forced to choose between binding to all interfaces (by leaving the address value empty) or only one specific interface.

This PR introduces support for specifying a comma-separated list of local addresses for the RTC service to listen on.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-64239

### DIFF
--- a/service/rtc/config.go
+++ b/service/rtc/config.go
@@ -13,11 +13,11 @@ import (
 
 type ServerConfig struct {
 	// ICEAddressUDP specifies the UDP address the RTC service should listen on.
-	ICEAddressUDP string `toml:"ice_address_udp"`
+	ICEAddressUDP ICEAddress `toml:"ice_address_udp"`
 	// ICEPortUDP specifies the UDP port the RTC service should listen to.
 	ICEPortUDP int `toml:"ice_port_udp"`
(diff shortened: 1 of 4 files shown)


service/rtc/config.go
```diff
@@ -343,3 +343,23 @@ func (s *ICEHostPortOverride) UnmarshalTOML(data interface{}) error {
 
 	return nil
 }
+
+type ICEAddress string
+
+func (a ICEAddress) Parse() []string {
+	return strings.Split(string(a), ",")
```
cpoile, line 350: split and trim too, just in case? (or does ParseIP handle that?)

cpoile (approved):
just a non-blocking question
nice and clean

PR #179, opened 2025-06-16
[MM-64239] Filter out candidates for unused local addresses

streamer45:
#### Summary

Even when listening on specific interfaces (i.e., through the `ice_address_udp` and `ice_address_tcp` settings), the ICE layer would generate candidates for all available local addresses. This is problematic for a few reasons, but mostly because we wouldn't listen to all of those IPs. We add the necessary changes to filter out anything we are not going to listen to.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-64239

### DIFF
--- a/service/rtc/server.go
+++ b/service/rtc/server.go
@@ -118,7 +118,7 @@ func (s *Server) Start() error {
 				s.localIPs = append(s.localIPs, ip)
 			}
 		}
-		s.log.Debug("rtc: ICE addresses set, overriding local IPs", mlog.Any("localIPs", s.localIPs))
+		s.log.Debug("rtc: ICE addresses set, only using filtered local IPs", mlog.Any("localIPs", s.localIPs))
 	}
 
 	if m, _ := s.cfg.ICEHostPortOverride.ParseMap(); len(m) > 0 {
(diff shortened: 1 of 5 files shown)


service/rtc/server.go
```diff
@@ -105,10 +106,25 @@ func (s *Server) Start() error {
...
+	// If listening on specific IPs, ensure the localIPs are filtered accordingly.
+	// We should discard anything that isn't appearing in either udpLocalIPs or tcpLocalIPs.
+	if len(udpLocalIPs) > 0 || len(tcpLocalIPs) > 0 {
+		s.localIPs = nil
+		for _, ip := range localIPs {
+			if slices.Contains(udpLocalIPs, ip.String()) || slices.Contains(tcpLocalIPs, ip.String()) {
+				s.localIPs = append(s.localIPs, ip)
+			}
+		}
+		s.log.Debug("rtc: ICE addresses set, overriding local IPs", mlog.Any("localIPs", s.localIPs))
```
cpoile, line 121: can we change to something like "rtc: ICE addresses set, filtered and only using local IPs"? (just to be a bit more clear when viewing logs)

cpoile (approved):

